PR #7, opened 2024-07-05
Basic CI pipeline

jrr:
- Configure a basic CI pipeline to run a build on every commit (based on [PlatformIO's example](https://docs.platformio.org/en/latest/integration/ci/github-actions.html#using-cmd-run-command))
  - You can see example job output on my fork [here](https://github.com/jrr/trmnl-firmware/actions/runs/9802709039/job/27067764667)
- Correct an `#include` that was broken on case-sensitive filesystems (like Linux)

Down the road unit tests could run in CI, too, when those exist ;)

Congrats on the Kickstarter success! I'm excited for the project.


ryanckulp:
thanks @jrr ! tagging Michael from our team to review.

PR #12, opened 2024-07-12
Update Espressif platform

jrr:
I noticed that the project was a couple versions behind on the Espressif platform. Probably want to test this on hardware before merging it :)


MichaelXlivnenko:
Good day, thanks for your comments, we are planning to upgrade to 6.7.0.  we are currently in the process of testing

MichaelXlivnenko:
All tested. Version will be changed in the next version.

PR #28, opened 2024-08-20
Added support for adding multiple WiFis

paw3lx:
- Added support for adding multiple WiF credentials - user will be able to add up to 5 wifi credentials
- Added autoconnect method - which connects to the wifi based on the "standard" method (I think that phones/laptops behaves like that).  So at first it tries to connect to the last used WiFi network; in case it fails, it scans available networks and connects to the strongest one.
- Removed old WiFiManager library


paw3lx:
image

ryanckulp:
looks great @paw3lx ! stand by for Michael to review.

PR #31, opened 2024-09-04
A few more tests

jrr:
More unit test coverage for API response parsing and BMP header parsing.


lib/trmnl/src/parse_response_api_display.cpp
```diff
@@ -16,11 +16,11 @@ ApiDisplayResponse parseResponse_apiDisplay(String &payload)
...
   String special_function_str = doc["special_function"];
 
-  return {
+  return ApiDisplayResponse{
       .status = doc["status"],
-      .image_url = doc["image_url"],
+      .image_url = doc["image_url"] | "",
       .update_firmware = doc["update_firmware"],
-      .firmware_url = doc["firmware_url"],
+      .firmware_url = doc["firmware_url"] | "",
```
jrr, line 23: The `| ""` ensures that absent fields are parsed into empty strings (instead of the string `"null"`)

PR #33, opened 2024-10-08
Using different API endpoint

ronakjain90:
The Idea behind this PR is separation of concern. Having a different endpoint than the mother app, lets us create dedicated server to serve devices 1000s of devices when they come online and request display data. This is more to future proof scalability when needed.


ronakjain90:
Hi @jrr, @MichaelXlivnenko - I created this PR, would really appreciate if you can review this feature request.

PR #38, opened 2024-10-10
Fixed several logging bugs

GoReNeY:
- Fixed bug when logs won't be cleared from preferences when successfully sent to the server
- Fixed bug with incorrect log arguments formatting

### DIFF
--- a/src/bl.cpp
+++ b/src/bl.cpp
@@ -65,8 +65,6 @@ static bool saveCurrentFileName(String &name);
 static bool checkCureentFileName(String &newName);
 int submitLog(const char *format, ...);
 
-#define submit_log(format, ...) submitLog("%d [%d]: " format, getTime(), __LINE__, ##__VA_ARGS__);
-
 void wait_for_serial()
 {
 #ifdef WAIT_FOR_SERIAL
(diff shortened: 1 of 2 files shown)


src/bl.cpp
```diff
@@ -1772,14 +1769,21 @@ static bool checkCureentFileName(String &newName)
 
 int submitLog(const char *format, ...)
 {
+
+  memset(log_array, 0, sizeof(log_array));
+  snprintf(log_array, sizeof(log_array), "%d [%d]: ", getTime(), __LINE__);
```
jrr, line 1774: I suspect the reported logs will all now say line `1774` 😐

jrr:
@GoReNeY can you describe the buggy behavior you were fixing? We ought to able to write testable functions that construct messages correctly without losing out on the line number macro.

GoReNeY:
> @GoReNeY can you describe the buggy behavior you were fixing? We ought to able to write testable functions that construct messages correctly without losing out on the line number macro.

Yeah, my bad, I was fixing a bug with logging arguments formatting and removed the macro in progress, I'm about to return it

PR #40, opened 2024-10-14
FIxed some build warnings

GoReNeY:
- Fixed redefinition of WIFI_CONNECTION_ATTEMPTS
- Fixed "control reaches end of non-void function" in filesystem_file_rename function
- Deleted unused "WIFI_PORTAL_TIMEOUT" define


GoReNeY:
There are also some warnings within ESPAsyncWebServer fork that we use, but it seems to be non-critical and may be fixed soon

PR #43, opened 2024-10-19
Version updates

jrr:
- update espressif32 to 6.9.0
- pin ESPAsyncWebServer to 3.3.0

I haven't tested this on a device yet, but will follow up when I get to it.


GoReNeY (approved):
Tested on the device, seems to be working fine

PR #60, opened 2025-02-04
Send width and height

deisterhold:
Add headers for display width and height so the server can dynamically adapt the generated image.

### DIFF
--- a/src/bl.cpp
+++ b/src/bl.cpp
@@ -553,6 +553,8 @@ static https_request_err_e downloadAndShow(const char *url)
     https.addHeader("Battery-Voltage", String(battery_voltage));
     https.addHeader("FW-Version", fw_version);
     https.addHeader("RSSI", String(WiFi.RSSI()));
+    https.addHeader("Width", String(EPD_7IN5_V2_WIDTH));
+    https.addHeader("Height", String(EPD_7IN5_V2_HEIGHT));
 
     Log.info("%s [%d]: Special function - %d\r\n", __FILE__, __LINE__, special_function);
     if (special_function != SF_NONE)


ryanckulp:
@deisterhold thank you. could you also go ahead and clean up `EPD_7IN5_V2_*` vars to simply `EPD_WIDTH`, `EPD_HEIGHT` throughout to make this even better? then we'll merge and make these values accessible inside the private plugin builder / merge variables.

deisterhold:
Those values are defined in the individual header files for each display model. How do you feel about me adding functions to the display class for returning these values instead? Then everything remains encapsulated.

ryanckulp:
@GoReNeY please make this decision ;)

GoReNeY:
> Those values are defined in the individual header files for each display model. How do you feel about me adding functions to the display class for returning these values instead? Then everything remains encapsulated.

Yup, sounds great

deisterhold:
Done.

GoReNeY:
> Done.

Thanks for your contribution!

danielsitnik:
I'm not an expert on the firmware but I'll just leave my 2 cents here.

I know the server (including BYOS implementations) also requires the bits per pixel (or bit depth) when rendering the image. If the point is making everything dynamic to support future devices, then shouldn't this information be sent as well?

I'm working on a BYOS implementation and right now my database has a table called `device_models` where I'm storing width/height/bits. And then each `device` registered is also linked to a `device_model` in this table so during image generation I can retrieve these parameters.

ryanckulp:
@danielsitnik server is now storing width/height headers (from this PR), and will soon make them accessible via global vars like ` {{ trmnl.device.width }}`.

danielsitnik:
@ryanckulp but this logic has been added to the call to `/api/display`, and not `/api/setup`. So what you mean is that everytime the device pings the server for a new image you're also updating the database with these same values over and over? Wouldn't it make more sense to do it just once during setup?

But regardless of setup or display, my point is that the bit depth information is as valuable as the width/height because it will be used during image generation...

signalwerk:
I can't see the headers in the request. Am I doing something wrong?

I tried to read the values from the headers in combination with the [Redirect Plugin](https://help.usetrmnl.com/en/articles/11035846-redirect-plugin)

On the requests to the JSON-Data I get the headers like this:

```
'user-agent': 'Ruby'
accept: 'application/json'
'accept-encoding': 'gzip;q=1.0,deflate;q=0.6,identity;q=0.3'
content_type: 'application/json'
```

and on the request for the bmp:

```
'user-agent': 'ESP32HTTPClient',
'accept-encoding': 'identity;q=1,chunked;q=0.1,*;q=0'
```

I was in particular interested also in `Battery-Voltage` but I also don't see that header.

ryanckulp:
@signalwerk not yet announced but we now have a `/api/devices` endpoint to fetch voltage, battery charged, wifi signal etc:
https://usetrmnl.com/api-docs/ 

(requires a User API Key from your account page)

signalwerk:
@ryanckulp that is great! so nice. I was actually able to do what i wanted with `/api/devices`. Thank you for pointing this out!

Issue #61, opened 2025-02-05
Option to choose self hosted server at initial login


kaiwei:
Instead of forking the firmware to change the base url, is it possible (via the UI) to choose and the input a self hosted server address - as an option? As such, the default firmware will always work for self hosting or the default cloud server out of the box. 

Something like this below:

https://images.app.goo.gl/xVb1W35YxUTC9QZUA

Omeryl:
I was actually going to work on this, once I got my physical device. Is there any appetite for this @ryanckulp? I’m fine with it even being a hidden page not linked from the default captive portal to make it so your main customers are not led astray!

gitstua:
This would be great @ryanckulp - we could also point at a preview cloud instance to help test new updates. As the userbase grows we want to figure out how to keep any markup changes stable for standard users and let dev hackers use a preview server.

ryanckulp:
thanks everyone, i really like this idea. for context, i initially wanted to provide a firmware config GUI here:
https://usetrmnl.com/flash

then users could mod multiple settings and flash directly to their device without installing any dependencies.

but the ESP Web Tools lib does not actually build files (AFAIK), just facilitates their upload to a peripheral. so it probably isn't worth running server-side builds and opening up that can of worms.

ergo! FW maintainer @GoReNeY -- any objections to making `API_BASE_URL` definable by users inside the WiFi captive portal? also tagging @paw3lx who did most of the [relevant work](https://github.com/usetrmnl/firmware/pulls?q=is%3Apr+is%3Aclosed+author%3Apaw3lx).

Omeryl:
Cool, I've been working on this and I'm nearly complete, I'm just testing it now and will PR it later today hopefully.

ryanckulp:
completed via #66

PR #66, opened 2025-02-12
Set API URL via Captive Portal Setup and allow http/https API URLs

Omeryl:
This PR introduces the ability to define a custom server in the wifi captive portal setup. It gives a fair warning before letting the user click around, I did thorough testing to ensure upgrades work as expected as well as new set ups. I also took the time to allow both HTTP/HTTPS connections depending on the api url.

Related issue: #61


Omeryl:
Gah, excuse the auto formatting from my VSCode instane.

lib/wificaptive/portal/index.html
```diff
@@ -338,12 +338,35 @@
...
+        } catch (_) {
+          displayAlert('Please enter a valid URL for your custom server', "warning");
+          return;
+        }
+      }
+
+      // strip trailing slashes
+      server = server.replace(/\/+$/, '');
+      if (!server) {
+        server = "https://trmnl.app";
```
ryanckulp, line 365: possible to leverage `API_BASE_URL` here instead of duplicating? not a deal breaker, but we may have to start rotating alias domains in the future so would like to keep as DRY as possible.

lib/wificaptive/src/WifiCaptive.h
```diff
@@ -36,62 +34,67 @@
...
+        int32_t rssi;
+        bool open;
+        bool saved;
+    };
+
+    DNSServer *_dnsServer;
+    AsyncWebServer *_server;
+    String _ssid = "";
+    String _password = "";
+    String _api_server = "https://trmnl.app";
```
ryanckulp, line 57: same comment as before, ideally point to our constant default

ryanckulp:
thank you @Omeryl !

@GoReNeY please add comments if applicable and let's get this baby merged. will make BYOS even easier, no FW mods or builds required. (thus no VS Code or PIO required either , great for us Atom text editor OGs). wish list: tag this + recent refactorings as v1.4.6.

lib/wificaptive/src/WifiCaptive.h
Omeryl: done!

Lukas-Heiligenbrunner:
Is there any way to reset this after setup is complete? 
Had to erase the flash to get rid of the settings^^

PR #68, opened 2025-02-17
Cleaned up submit_log.cpp + removed memory leak

bdejong:
There was a memory leak in `submit_log.cpp` and the code was a little hard to follow. I cleaned up the memory leak and used the "early return" pattern to make the code easier to grok. I hope RAII is acceptable in the firmware...

Caveat: this is **untested** as I do not have a TRMNL device to test it on

### DIFF
--- a/src/api-client/submit_log.cpp
+++ b/src/api-client/submit_log.cpp
@@ -15,9 +15,9 @@ bool submitLogToApi(LogApiInput &input, const char *api_url)
   
   // make_unique might throw bad_alloc if out of memory
   if (isHttps)
-    client = std::make_unique<WiFiClientSecure>();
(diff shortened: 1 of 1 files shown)


Omeryl:
Working on reviewing and testing this now!

Omeryl:
Hey @bdejong,

Was trying to get this tested and integrated before the next firmware release, but for an HTTPS url it just complains about being unable to connect (-1). The same URL works just fine against current `main` branch, so something here feels particularly odd.

Omeryl (changes requested):
but for an HTTPS url it just complains about being unable to connect (-1). The same URL works just fine against current main branch, so something here feels particularly odd

PR #70, opened 2025-02-20
Estimation of work needed to support 1 bit PNGs as payload

delhoume:
this PR is not intended to be merged but provides information on what changes supporting 1 bit PNGs as a payload in addition to BMP must be done.

The idea is to keep "buffer" that will hold the final bitmap but there is another buffer for http payload and  deflate decompression (see top of bl.cpp).
This allows to  decorelate displayed bitmap and downloaded one that can be a BMP or a PNG.

BMP are decoded to "buffer"  using memcpy while PNG go through dynamic allocation and various stages to end up as the final displayed BMP.

there is enough memory on the esp32   to have two 48000  bytes buffers ( and some usually much smaller for compressed PNG data)

There are a number of things to be aware of, for example uncompressed PNG data is not exactly like final BMP, because of a 1 byte additional per row filter info. so decoded PNGs are 101x480 bytes !
Another one is that padding is at multiple of 4 bytes for BMPs but 1 byte for PNGs ! In TRMNL case they are the same, with a 100 byte width...
Another one is that by default BMPs are bottom-top but a negative height indicated top-bottom data...of corse PNGs are always top-bottom so they have to be inverted vertically...

I handle endianness in a generic way, I do not think it is worth having native conversion for esp32 because of the very low number of endian-specific int16 and int32 in BMPs or PNGs

miniz is used to (de)compress data, but not as a library dependency.

the main code is in onebit_bmp_png.h with support from onebit,h

As I do not have yet a TRMNL device (because of the sell-out), this code is untested and will likely crash.
I went as far as having no compilation errors but I am confident there are bugs...

The conclusion is that not much work is needed for minimal support, it is mostly limiting "buffer"  to the displayed bitmap and have different one(s) for http payload or temporary data. 

If any clarification is wanted on the code or workflow, feel free to ask.


ruohki:
In my byod device iam using png. I made myself a web proxy that converts the bmp images to png on the fly. My firmware uses esphome.

delhoume:
> In my byod device iam using png. I made myself a web proxy that converts the bmp images to png on the fly. My firmware uses esphome.
> 
> https://github.com/esphome/esphome/blob/dev/esphome/components/online_image/png_image.cpp

In the TRMNL context, only  1 bit BMP and PNG are supported, whereas I believe your code handles arbitrary images.

1 bit PNGs with typical TMNL UI framework generated content can be 1/10 of the size of the same image as a 1 bit BMP, saving bandwidth and requiring less time between device sleep

ruohki:
> > In my byod device iam using png. I made myself a web proxy that converts the bmp images to png on the fly. My firmware uses esphome.
> > https://github.com/esphome/esphome/blob/dev/esphome/components/online_image/png_image.cpp
> 
> In the TRMNL context, only 1 bit BMP and PNG are supported, whereas I believe your code handles arbitrary images.
> 
> 1 bit PNGs with typical TMNL UI framework generated content can be 1/10 of the size of the same image as a 1 bit BMP, saving bandwidth and requiring less time between device sleep

Those PNGs are smaller than the bitmaps. The display component does support BINARY type of images so 1-bit you gotta dig the code

Fr3dr1ckson:
Implemented in v1.5.0:
c2f210d88cfd3d721ff5437d42f8ac6412885ac3
c4a82c2ad947b88ae5792a310de5b73a141c566f

PR #72, opened 2025-03-01
fix: JSON warnings

beanieboi:
This fixes the following warning:

`fix warning: 'DynamicJsonDocument' is deprecated: use JsonDocument instead [-Wdeprecated-declarations]`


beanieboi:
closing this as #73 is a better solution

PR #75, opened 2025-03-06
Enable Longer Sleep Times

matthewlyons:
The current calculations of using a 32 bit int has an overflow when the server gives it a refresh_rate above 4294 seconds (about 70 minutes). 

Switching this to a 64bit integer will allow users to have their devices refresh once a day if they would like.


Omeryl (approved):

ryanckulp:
thank you @matthewlyons !

Issue #71, opened 2025-02-24
Multiple WiFi network credentials


ryanckulp:
i thought we already supported this, but maybe not.

some users transport their TRMNL between work <> home and expect it to automatically connect to the appropriate network, like other devices.

let's refactor [wifi credentials](https://github.com/search?q=repo%3Ausetrmnl%2Ffirmware%20ssid&type=code) to an array/collection that cycles each known network pairing before giving up and exposing the captive portal. and on boot, the last connected network should be attempted first before cycling.

michaelbolanos:
Here's a Python script that manages Wi-Fi connections by cycling through known networks before exposing a captive portal. It attempts to reconnect to the last successful network first. This is designed for Linux-based systems using nmcli (NetworkManager CLI).

Features:
Stores Wi-Fi credentials in an array.
Attempts to reconnect to the last successful network first.
Cycles through all known networks if the first attempt fails.
If all connections fail, it exposes a captive portal.

[wifi-refactor-rename-to-py.txt](https://github.com/user-attachments/files/18945874/wifi-refactor-rename-to-py.txt)

Omeryl:
@michaelbolanos I'm rather confused how that is relevant here.

michaelbolanos:
> [@michaelbolanos](https://github.com/michaelbolanos) I'm rather confused how that is relevant here.

It was a mistake.

gdifiore:
I'm taking a crack at this

delhoume:
WifiCredential already stores an array of SSIDs that had a successful connection to.
And looks like they are automatically sorted by latest date.
So instead of giving up  if _lastUsed fails at line 190 wificaptive.cpp, you just have to try next one in the array (loop over _savedWifis)

Maybe 5  to 10 lines of code

gdifiore:
@delhoume Yes, that's exactly what I wrote on my local branch, just waiting for my device to come this week before testing and pushing PR.

delhoume:
> [@delhoume](https://github.com/delhoume) Yes, that's exactly what I wrote on my local branch, just waiting for my device to come this week before testing and pushing PR.

There might be a bug in saving credentials, the ssid and password are not checked for existence in the array so maybe you will have multiple times the same  pair in _saveWifis and nothing but this pair...

And maybe I would  make readWifiCredentials take an index an index so that it can be looped over.
and I would get  rid of _lastUsedXX, replaced by index 0.

[...]

PR #76, opened 2025-03-06
Multiple WiFi Network Support

gdifiore:
Issue #71
- Remove `_lastUsed`, will shift all network credentials in the array so that last used wifi is `_savedWifis[0]`
- On failed connection to last used WiFi network, loop over all saved credentials in the array
- No longer saves duplicate networks in the array

### DIFF
--- a/lib/wificaptive/src/WifiCaptive.cpp
+++ b/lib/wificaptive/src/WifiCaptive.cpp
@@ -314,22 +314,15 @@ bool WifiCaptive::isSaved()
     return _savedWifis[0].ssid != "";
 }
 
-void WifiCaptive::readWifiCredentials(int index)
+void WifiCaptive::readWifiCredentials()
 {
     Preferences preferences;
     preferences.begin("wificaptive", true);
 
-    if (index >= 0 && index < WIFI_MAX_SAVED_CREDS) {
-        // Read only the specified index
-        _savedWifis[index].ssid = preferences.getString(WIFI_SSID_KEY(index), "");
-        _savedWifis[index].pswd = preferences.getString(WIFI_PSWD_KEY(index), "");
-    } else {
-        // Read all credentials (original behavior)
-        for (int i = 0; i < WIFI_MAX_SAVED_CREDS; i++)
-        {
-            _savedWifis[i].ssid = preferences.getString(WIFI_SSID_KEY(i), "");
-            _savedWifis[i].pswd = preferences.getString(WIFI_PSWD_KEY(i), "");
-        }
+    for (int i = 0; i < WIFI_MAX_SAVED_CREDS; i++)
+    {
+        _savedWifis[i].ssid = preferences.getString(WIFI_SSID_KEY(i), "");
+        _savedWifis[i].pswd = preferences.getString(WIFI_PSWD_KEY(i), "");
     }
 
     preferences.end();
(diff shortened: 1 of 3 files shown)


lib/wificaptive/src/WifiCaptive.cpp
```diff
@@ -317,35 +314,46 @@ bool WifiCaptive::isSaved()
     return _savedWifis[0].ssid != "";
 }
 
-void WifiCaptive::readWifiCredentials()
+void WifiCaptive::readWifiCredentials(int index)
```
gdifiore, line 317: @delhoume let me know if this is what you meant in your [comment](https://github.com/usetrmnl/firmware/issues/71#issuecomment-2692851881). Not sure what the benefit of getting an individual index in the array is

gdifiore: If you wanted it to return the creds from a given index, it should probably be its own function. However, the array max size is so small, it's probably fine leaving functionality as iterating over the entire array (line 542)

delhoume: yes I wanted to be able to iterate the array from outside.
but as the coder you have the liberty and prerogative to do as you like.
if the initial issue that was the ability to connect automatically to previous connected SSIDs is done then its ok for me

gdifiore: I removed the indexing because I didn't use it in this PR

Omeryl:
I'm reviewing this commit, have a couple things I've discovered (some of which was bad behavior from before this commit that I'm working around now!) More momentarily.

Omeryl:
OK, so I fixed two things:

1. The first commit fixes a weird race condition where the library will return -2 (SCAN_FAILED) for a scan just before populating the data for the scan. This appears to be so that they can prevent deadlocks, but because we're not using the event loop to check for the async response it just causes endless rescanning if you're in an area with a ton of SSIDs (like what I have). I changed the function to use the synchronous call, and also added code in case someone swapped it back to async.
2. I noticed that you didn't reimplement the last used WiFi reconnection feature, so I'd constantly always try the first one saved every time. This is rather inefficient, so I created some functions to bring back the previous behavior (by storing the last index of a successful ssid/pw combo). 

If you could give these a quick try again @gdifiore I'd appreciate it! If it looks good to you as well, I'd say this is good to merge.

gdifiore:
1. I noticed that as well, but assumed I did something wrong on the connection v.s. a code issue. Thanks for fixing!
2. It does connect to last used WiFi, I just removed the specific variable for it. When it adds a new WiFi it shifts the array one over (removing the oldest entry) then adds the new credentials pair into the 0 index. 

I believe with your update if a user fills the credentials array and the most recently used is the last index, adding a new one would removed the previous most recently used. Just a UX issue

Omeryl:
For number 2, I specifically tested it for a good hour and it always went back to the last added WiFi network no matter the status. If that WiFi network is bad, it _should_ try the last successful one. That's what my second change fixes :)

Omeryl:
@gdifiore I fixed the issue with shifting credentials breaking the true last index :)

Omeryl (approved):

gdifiore:
Ahh yeah last-used-order is different from last-added order. Thanks for catching that!

Tested it again, looks good to me.

Omeryl:
@ryanckulp Ready to merge!

PR #77, opened 2025-03-06
Support for EPDIY Hardware

Lukas-Heiligenbrunner:
EPDIY is a open hardware project for epaper displays:
https://github.com/vroland/epdiy

I've ported your firmware to this platform. 
I tried to be as non-intrusive as possible.
All of the changes are only compiled if you target the epdiy platformio target.
Display size + display type are configureable with an env variable.
Text rendering is still missing and I'll add this later.

I've used this for the last several weeks and it works great so far.
Please review my code. If you guys are not interested in this I'll continue it in a soft-fork. :)

And should i add documentation somewhere? ;)

I've emailed @ryanckulp if you guys are interested in porting this back.
Thanks for your time.
-Lukas


Lukas-Heiligenbrunner:
Sry, overlooked the var name change. should be fine now. :)

src/display.cpp
```diff
@@ -63,13 +118,47 @@ uint16_t display_width()
  */
 void display_show_image(uint8_t *image_buffer, bool reverse)
 {
-    auto width = display_width();
-    auto height = display_height();
+    uint16_t width = display_width();
+    uint16_t height = display_height();
+
+#ifdef EPDIY
```
furkanmustafa, line 124:
```suggestion
#if IMAGE_BPP == 1 && DISPLAY_BPP == 4
```

src/display.cpp
```diff
@@ -63,13 +118,47 @@ uint16_t display_width()
...
+    epd_poweron();
+    epd_fullclear(&hl, temperature);
+
+    epd_copy_to_framebuffer(paint_area, image_buffer_4bpp, epd_hl_get_framebuffer(&hl));
+
+    enum EpdDrawError _err = epd_hl_update_screen(&hl, MODE_EPDIY_WHITE_TO_GL16, temperature);
+    Log.info("%s [%d]: Paint_NewImage %s\r\n", __FILE__, __LINE__, _err);
+    delete[] image_buffer_4bpp;
+    epd_poweroff();
+#else
```
furkanmustafa, line 156:
```suggestion
#elseif IMAGE_BPP != DISPLAY_BPP
#error Unsupported display configuration for BPP conversion
#else
```

src/display.cpp
```diff
@@ -111,6 +202,8 @@ void display_show_image(uint8_t *image_buffer, bool reverse)
  */
 void display_show_msg(uint8_t *image_buffer, MSG message_type)
 {
+#ifdef EPDIY
+#else
```
furkanmustafa, line 206:
```suggestion
#ifndef EPDIY
```

furkanmustafa (approved):
I have made a basic review (just by reading code, I haven't tested anything, including my suggestions).

It seems to me that this PR achieves non-intrusion of defaults perfectly, when added, it would not break or change any current behavior with TRMNL devices. It can be merged as-is. :heavy_check_mark: 

My (partly incomplete) suggestions point towards a direction of more generalization of arbitrary device support, with having TRMNL specifics as defaults everywhere possible.

And if possible, the following would be even better;
* much less mentions of `#ifdef EPDIY`
* use something like `DEVICE_TYPE=EPDIY` instead of `EPDIY`
* use feature toggling instead, wherever possible, like `#if PIN_INTERRUPT >= 0`
  * (and use `PIN_INTERRUPT=-1` to disable it for a device configuration)
* define even more system specifics as device configuration variables,
  * like `ADC_DIVIDER=64.000` (`/128)*2`) (and `ADC_DIVIDER=91.2981` (`/128)*1.402)`) for EPDIY)
* Allow downloading display native bpp images, and skip bpp conversion if the remote image is like that.

PS. My review itself might require a review.

bitbank2:
I'm not sure why this is still open. EPDIY code is not needed in trmnl; all parallel eink panels use FastEPD.

PR #85, opened 2025-03-08
Convert algo block scheme to dot format

datacompboy:
To make it easier for making changes in a tracked way, I find it easier to have them in text form.

It's less pretty in terms of WYSIWYG but it is nice in WYSIWYM paradigm and allow edits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,115 @@ created for the [TRMNL](https://usetrmnl.com) e-ink display.
 
 ## **Algorithm block scheme**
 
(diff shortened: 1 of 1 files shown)


datacompboy:
You can see the github-rendered view formatted here: https://github.com/usetrmnl/firmware/blob/48fed8f965ea44871caea477a299c8b6154d1d70/README.md

ryanckulp:
nice work @datacompboy. we may want to put it inside a new dir + link from README vs embed, just to prevent scroll fatigue. but looking good.

PR #87, opened 2025-03-11
Increase timeout for HTTPClient to 15 seconds during downloadAndShow

### LINKED ISSUE #82: Feature Request: Increase timeout when calling `/api/display`
My BYOS generates screens on the fly, and because I'm hosting on a cheap fly.io instance one of my screens can sometimes take 5-6 seconds to generate (it has to make 40-ish HTTP requests and then render some SVG).

This seems to be on the limit of the TRMNL's current timeout setting, as sometimes I get this in my logs:
```
  "log_message": "HTTP Client failed with error: read Timeout",
  "log_codeline": 1071,
  "log_sourcefile": "src/bl.cpp",
```

It would be great if the timeout could be increased to something more like 10s? Or even 15? I appreciate it shouldn't be too long, but a little more breathing room would be appreciated.

jamesthurley:
I believe this should resolve #82. Sets the timeout to 15 seconds.

I've chosen 15 seconds to try and get a balance between not waiting too long, but also giving plenty of time for dynamically generated content to be generated on cheap servers.

The reasoning is that the Images are going to be displayed on the device for quite a while so waiting a few extra seconds for them when fetching isn't an issue. By increasing the timeout it allows users to use cheaper, slower servers to generate their TRMNL screens.

### DIFF
--- a/src/bl.cpp
+++ b/src/bl.cpp
@@ -500,6 +500,10 @@ static https_request_err_e downloadAndShow()
   { // Add a scoping block for HTTPClient https to make sure it is destroyed before WiFiClientSecure *client is
 
     HTTPClient https;
+
+    // Set a more generous timeout than the default of 5000.
+    https.setTimeout(15000);
+
     Log.info("%s [%d]: RSSI: %d\r\n", __FILE__, __LINE__, WiFi.RSSI());
     Log.info("%s [%d]: [HTTPS] begin /api/display/ ...\r\n", __FILE__, __LINE__);
     char new_url[200];


ryanckulp:
thanks for this. just FYI, on core we generate screens in a separate process, so the `api/display` endpoint always responds immediately with prepared content. 

so while this makes sense for BYOS, i'm wondering if this should instead become a user-facing config. it could be a new flag in the `api/display` response payload, or even a UI form field in the FW's captive portal display (a la #66). my concern is draining battery from core-pinging devices that should probably just give up after the default 5000ms.

jamesthurley:
Yes, fair enough. I'll work on making it configurable via the response payload.

jamesthurley:
@ryanckulp Changes made so now the server can optionally return a `image_url_timeout` alongside the `image_url` in the `/api/display` payload. If the server doesn't populate this then the behaviour is exactly the same as before, using the default timeout.

Side note:
The HTTPClient only supports a UINT16 timeout, so about 65 seconds max. If anyone did ask for more, which is unlikely but not completely implausible, I tried to handle it gracefully along with sending a log message back to the server. Hopefully that will avoid confused support questions about long timeouts not being respected.

ryanckulp:
thanks @jamesthurley . i can get this `image_url_timeout` added to the server response, but it's not clear what happens if the key* exists but value is either `nil` or `""` (empty string). will that still coerce to `0` in the check that falls back to default? and i suppose this value should be in ms? so worst case we can have the server always* reply `500`?

jamesthurley:
@ryanckulp @Omeryl Do you think it's an issue that I put the `image_url_timeout` in milliseconds when `refresh_rate` is in seconds? I didn't notice the inconsistency at the time.

On one hand the inconsistency bothers me, on the other hand perhaps hand timeouts are normally in milliseconds so it's expected. 🤔

Anyway, as likely no-one is using this feature yet except me, I could create a new PR that changes it to seconds if you think that would be better.

jamesthurley:
Hmm, ok, so the question is, do we (or TRMNL) want to make this consistent? I can think of some options if so.

bkuhlmann:
You have my vote for making this consistent (I'm with you there). The hard part is using seconds or milliseconds because milliseconds for refresh rate isn't exactly friendly math but sounds like seconds for image timeout URL would not be granular enough?

I suppose all values could be in milliseconds and, for the refresh rate, milliseconds would always be encoded/decoded into seconds for UI purposes only. I could see that being a potentially working if others agree.

jamesthurley:
I'm thinking that we shouldn't change `refresh_rate`, as it would create a big backwards compatibility headache.

Having the `image_url_timeout` in second is fine I think... in the context of a TRMNL device it shouldn't need any more granularity.

I've created a PR with some notes here: https://github.com/usetrmnl/firmware/pull/100

PR #88, opened 2025-03-11
Add ID header to /api/log call

### LINKED ISSUE #80: Include device ID header with `/api/log` calls
Currently the access token is included, but not the ID:

https://github.com/usetrmnl/firmware/blob/aa476cbb6db0cff32cc78243e18d347c984a8998/src/api-client/submit_log.cpp#L41-L43

Including the ID would allow BYOS to only show logs from authenticated devices.

jamesthurley:
This adds the ID header to the `/api/log` call.

Reasoning:
- It improves consistency, as both `/api/setup` and `/api/display` currently receive the device ID.
- It allows you to identify a device which has a bad or expired API key from `/api/log` calls.
- It allows looking up a device by ID on all API calls.

Resolves #80

Tested on device.

### DIFF
--- a/src/api-client/submit_log.cpp
+++ b/src/api-client/submit_log.cpp
@@ -38,6 +38,7 @@ bool submitLogToApi(LogApiInput &input, const char *api_url)
       { // HTTPS
         Log_info("[HTTPS] POST...");
 
+        https.addHeader("ID", WiFi.macAddress());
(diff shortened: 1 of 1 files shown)


Omeryl:
@ryanckulp This looks good to me.

Omeryl (approved):

PR #89, opened 2025-03-11
Add support for ETag header

### LINKED ISSUE #83: ETag support
I think that `usetrmnl.com` will compare the newly generate image with the previously generated image, and if they are the same it won't send a new image to the TRMNL device, saving bandwidth and battery.

In my BYOS I generate screens on the fly and don't cache them anywhere, so I can't compare images, but supporting [ETags](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) in the firmware would allow servers to implement the same optimization without caching images. It could potentially be a more efficient way for `usetrmnl.com` to do it as well.

The server would send the image hash back as an ETag header along with the image, and the TRMNL would store the ETag and send it back to the API in the headers of the next `/api/display` request. The server can then simply compare ETags to determine if the screen has changed.

jamesthurley:
Resolves #83.

This commit adds ETag support when requesting the `image_url` returned by the `/api/display` call.

- If an ETag was previously received it is added to the request headers.
- If the server returns 304 Not Modified, it doesn't fetch the body or refresh the display.
- If the server returns an ETag then it is saved to include in the next `image_url` request.
- If the logo is displayed on the screen (due to refreshing via the back button) the ETag is cleared to ensure we draw the latest returned image.

If the server doesn't support ETags then the there are no significant changes in behaviour.

Tested on my device.

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -22,6 +22,7 @@
 #define PREFERENCES_LAST_SLEEP_TIME "last_sleep"
 #define PREFERENCES_CONNECT_API_RETRY_COUNT "retry_count"
 #define PREFERENCES_CONNECT_WIFI_RETRY_COUNT "wifi_retry"
+#define PREFERENCES_ETAG "etag"
 
 #define WIFI_CONNECTION_RSSI (-100)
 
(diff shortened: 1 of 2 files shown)


schrockwell:
Thanks for this PR, but I am going to close it for now. A lot has changed in the past year, and this optimization is no longer necessary. 

The canonical way to detect a new image with the TRMNL API is to check if the filename has changed. The filename itself functions as the cache key. Adding ETag support would introduce a second layer of caching, and then it would be difficult for users and BYOS developers to debug unexpected caching behavior.

This filename-check optimization was introduced in 1b6dd33 and improved in #479. I understand that ETags are the more Web-friendly mechanism to solve this problem, but the extra caching complexity is unfortunately not worth it when a solution already exists.

Issue #86, opened 2025-03-10
Remove HTTP status from responses


bkuhlmann:
## Why

Would it be possible to remove the HTTP status from API responses? This would allow us to reduce duplication since we'll be able to acquire this information from the HTTP status code.

## How

At the moment, our API responses look like this:

<details>
  <summary>`/api/display`</summary>

``` ruby
{
  filename: "test",
  firmware_url: nil,
  image_url: "/assets/images/generated/test.bmp",
  refresh_rate: 900,
  reset_firmware: false,
  special_function: "sleep",
  status: 200,
  update_firmware: false
}
```
</details>

<details>
  <summary>`/api/setup`</summary>

``` ruby
{
  status: 200,
  api_key: "abc123,
  friendly_id: "ABC",
  image_url: "/assets/images/setup/logo.bmp",
  message: "Welcome to TRMNL BYOS"
}
```
</details>

I'd like to remove from `status` key from the above responses so you can only rely on the HTTP status code instead. This would reduce payload size and duplication.

ryanckulp:
if you check out core > `api/display`, 

```rb
status = @device.user_id? ? 0 : 202 # forces device to render Setup screen until user is attached
```

this is why the status field is important. but in theory it could be migrated to `device_setup_at`, which could either be `nil` or a timestamp, to achieve the same thing.

Omeryl:
I'd love to see the firmware actually standardize on HTTP response codes myself, it's a bit less confusing.

bkuhlmann:
*Ryan*: The use of `0` is a bit strange too because it's not a valid HTTP status code...but a `setup_at` or `registered_at` would make a lot more sense. I like that a lot more. Would be great to fold that in and I can updated Core and BYOS.

*Elizabeth*: Yeah, the double notation of HTTP status code _and_ status within the payload would be nice to reduce because I'd much rather rely on the HTTP status code as intended by most APIs and then not have duplicate this in the response body. Less chance of confusion if the codes get out of sync between HTTP code and body too.

ryanckulp:
10-4. just got us started here:
https://github.com/usetrmnl/core/pull/791

could screenshot those todo's or cross post here in case @Omeryl wants to get involved. this will require a synchronous effort i think, along with releasing a new FW, unless @bkuhlmann prefers we maintain backward compatibility in our core stack for a few days.

Omeryl:
[...]

PR #91, opened 2025-03-13
add fw-version header to api/setup request

Omeryl:
Allows for us to standardize on the HTTP response code rather status "status" field in JSON, while maintaining backwards compat with older firmware.

Tested on device.


Omeryl:
This relates to #86.

danielsitnik:
I still don't understand why don't you add the width/height/bitdepth to this call instead of the /display one. 🤷‍♂️

Omeryl:
@danielsitnik I just added what was the bare minimum to get a feature they're trying to work on internally, if you want that data just open a PR and I doubt anybody would strike it down :)

danielsitnik:
Sure. I'm just mentioning it because it has been discussed before.

Not saying it should have been implemented by you. This is more to call the team's attention ;)

PR #94, opened 2025-03-16
Add offline clock display

datacompboy:
Requires /api/display/ to send clock_settings in response.

As long as the device is set up, configured time and have clock settings set up, the clock will be updated every minute (so no point in using :%S in format), without need for wifi connection. It won't change frequency of wifi connection, and only clears the clock area to reduce power consumption.

The clock settings are:
   coordinates of box (X/Y/X+W/Y+W)
   colors for text and background
   font size
   format string

If font size is zero or width or height of the box are zero, clock update is not enabled and thus won't wake up or draw anything.


ronakjain90:
Interesting, I assume that this will also work with mashups out of the box (just need x, y coordinates). So in theory, one could have a Mashup of 4 plugins, and the firmware can refresh just the clock area. TRMNL OG however, doesn't have partial screen refresh capability, so I wonder how did you only get this to clear just the cock x,y area.

Can the official clock be supported as well? the server can pre-generate 720 images (12 hrs * 60 mins) and the device can enumerate with file names?

This is an interesting approach to Clock implementation, as it's amongst our top 10 plugins.

datacompboy:
Yes, the clock itself is not dependent on any plugin at all: the idea was that clock can always be seen in one of the corners, where it won't interfere with useful data from real plugins.

For eInk displays, "clear" is typically output of one color then another, which is working like clear, so clearing window is possible. The library already exposed `Paint_ClearWindows` function for that. As I said, I haven't yet tested on the screen, maybe two calls needed (first with foreground then with background color) to wipe-clean clearing :).

While showing one-of-N images is possible, it will be quite expensive battery-wise, as there is not enough memory to store all the possible clock faces in the RAM nor ROM -- so you'll need to connect to wifi to show it every minute (well, maybe every 5.. if fits few in memory in compressed form?). But I would expect that it will cause wearout of the flash quite fast, withing few months...

datacompboy:
Yeah, getting an actual partial update not mess up with the screen is way harder than in theory. These screens are not what I expected...

jonm01:
I'm excited! Does it work with the trmnl device's factory display?

PR #98, opened 2025-03-26
Update display.cpp: fix use of the wrong sleep function

datacompboy:
7IN5B_V2 while everything else using 7IN5_V2

### DIFF
--- a/src/display.cpp
+++ b/src/display.cpp
@@ -336,5 +336,5 @@ void display_show_msg(uint8_t *image_buffer, MSG message_type, String friendly_i
 void display_sleep(void)
 {
     Log.info("%s [%d]: Goto Sleep...\r\n", __FILE__, __LINE__);
-    EPD_7IN5B_V2_Sleep();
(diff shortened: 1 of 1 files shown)


ryanckulp:
@yankobogdan does this look right to you? i have a vague memory of B_V2 versus V2 during prototyping stage.

datacompboy:
Well, unless you modify it, they are identical. Just I had issues because for my setup I had to edit it... 😂

delhoume:
While you are thee why not replace
#include "EPD.h"
with
#include "Debug.h"
#include "utility/EPD_7in5_V2.h"


in display.cpp

I see no point in including all possible panels...

datacompboy:
(rebased over latest main)

datacompboy:
> I see no point in including all possible panels...

Agree here, but that's story of another change with little refactoring. I delay that until multiple refactorings got merged to avoid duplicated work and many conflicts :)

datacompboy:
This is obsolete as eInk driver was replaced :)

PR #101, opened 2025-04-03
PoC BLE WiFi Provisioning

gdifiore:
- PoC implementation replaces captive portal
- Refactored common Wifi functionality out of WifiCaptive and into new WifiCommon
- Can be tested by:
    - [Wi-Fi BLE Tool](https://apps.apple.com/sa/app/wi-fi-ble-tool/id1553059224) on iOS
    - A [Minimal Web Interface](https://gist.github.com/gdifiore/ecfc2027437cf633814b4a49734b94de) on a PC


ryanckulp:
if/when we add BLE support it will be for screens themselves to transmit via BLE, not just WiFi credential sharing. sorry for not making this decision (much) sooner.

PR #102, opened 2025-04-05
Mac and Windows CI

jrr:
This PR:
 - ~~makes a couple small changes in platform.ini to allow `pio test -e native` to work with MacOS' current C/++ compiler.~~
 - adds CI jobs to build and test on MacOS and Windows.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,7 +4,15 @@ on: [push, pull_request]
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      max-parallel: 3
+      matrix:
+        os: [ubuntu-latest, macos-15, windows-latest]
+
+    runs-on: ${{ matrix.os }}
+
+    timeout-minutes: 10
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


Omeryl:
I'm not too sure the value in doing all 3 platforms, where we're really only concerned for doing the CI on one platform to be sure they pass. This could blow through CI credits pretty quickly for the organization.

jrr:
> I'm not too sure the value in doing all 3 platforms, where we're really only concerned for doing the CI on one platform to be sure they pass. 

The value would be in ensuring that the development environment continues to function on all platforms that a contributor may use. So, for example, to prevent a developer who uses Linux locally from inadvertently breaking the Windows build. I'm not sure how likely this is in PlatformIO projects, but I know there's at least a little bit of OS-specific behavior in the "native" test platform used for unit tests. (one such breakage on MacOS is what let me to do this)

> This could blow through CI credits pretty quickly for the organization.

That's a good point! GitHub's docs say it's [free for standard GitHub-hosted runners in public repositories](https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-actions/about-billing-for-github-actions#about-billing-for-github-actions), but I imagine there's an unstated limit _somewhere_. If their minute usage became a problem then the jobs could be configured to run less frequently. (say, only on PRs, instead of every push)

> Is there a way we could enforce a 5-10 minute timeout?

[yep!](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes). It looks like timeouts can be specified per-step, too, like just on the pio commands if we don't trust 'em.

Omeryl:
Cool! If we add some timeouts to the steps that resolves all my concerns, thanks @jrr!

Omeryl:
this looks good to me!

PR #108, opened 2025-04-11
Small refactor in downloadAndShow()

jrr:
I hope to gradually overhaul `downloadAndShow()`. But to make it easier to review and test, it will take a bunch of small steps to get there. Here's the first one, slicing off a few separate functions related to the `/api/display` call:
 - `loadApiDisplayInputs()` to load information out of preferences etc.
 - `addHeaders()` to take this information and tack on the headers.
 - `handleApiDisplayResponse()` to handle the (many) side effects afterward.

### DIFF
--- a/src/bl.cpp
+++ b/src/bl.cpp
@@ -479,6 +479,60 @@ void bl_process(void)
 {
 }
 
+ApiDisplayInputs loadApiDisplayInputs(Preferences &preferences)
+{
+  ApiDisplayInputs inputs;
+
+  inputs.baseUrl = preferences.getString(PREFERENCES_API_URL, API_BASE_URL);
+
+  if (preferences.isKey(PREFERENCES_API_KEY))
+  {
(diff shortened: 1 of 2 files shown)


src/bl.cpp
```diff
@@ -622,770 +662,790 @@ static https_request_err_e downloadAndShow()
...
+    send_log = false;
+  }
+  Log.info("%s [%d]: Returned result - %d\r\n", __FILE__, __LINE__, result);
+  return result;
+}
+
+https_request_err_e handleApiDisplayResponse(ApiDisplayResponse &apiResponse)
+{
+  https_request_err_e result = HTTPS_NO_ERR;
+
```
jrr, line 948: All the code below is pasted directly from downloadAndShow. It was formatted by my editor but the logic is unchanged.

jrr:
Rebased and tested on my device. I kept the commits tidy; it may be easier to review them one-at-a-time.

PR #111, opened 2025-04-12
Update espressif32 6.9 -> 6.10

jrr:
Noticed we were one rev out of date on the espressif framework.

It looks like there were no significant changes in [6.10](https://github.com/platformio/platform-espressif32/releases/tag/v6.10.0) (and it uses the same version of Arduino and IDF as [6.9](https://github.com/platformio/platform-espressif32/releases/tag/v6.9.0)). Tested on my device and it seems to work just fine.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -12,7 +12,7 @@
 default_envs = esp32-c3-devkitc-02
 
 [env:esp32-c3-devkitc-02]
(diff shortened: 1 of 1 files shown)


Fr3dr1ckson:
Implemented in a85006d5911f3985c4c68c1f97eae26107795788

PR #113, opened 2025-04-16
Scalable contents (SVG like pathes support)

delhoume:
This contribution does not change in any way the current implementation.

It shows how we can leverage external librairies to add features in a simple way.

This one uses a subset of gfxDraw (https://github.com/mathertel/gfxDraw) to display SVG  syntax vectorial objects).

It uses Waveshare's GUI_Paint a a backend.

The code is triggered with
#include <draw_scalable.h>

this brings a generic function
```
void draw_scalable(const char *svgPath, int16_t pathX, int16_t pathY,
                   int16_t pathWidth, uint16_t pathHeight, int16_t centerX,
                   int16_t centerY, int16_t width, UWORD fillColor = BLACK, UWORD strokeColor = WHITE, float rotation = 0.0);
```
and 3 TRMNL logo ones.

Sample usage in display.cpp for message screen (for example when asking for Wifi credentials):

```
...
    Paint_SelectImage(BlackImage);

#if #defined(SVG_LOGO)
  Paint_Clear(WHITE);
   const char* heart = "M43 7 a1 1 0 00-36 36l36 36 36-36a1 1 0 00-36-36z";
   draw_scalable(heart, 0, 0, 108, 82, 400, 150, 100, BLACK, BLACK, 0.0);
  draw_scalable(heart, 0, 0, 108, 82, 400, 155, 90, WHITE, BLACK, 0.0);
  drawFullLogo(400, 240, 384);
#else
  Paint_DrawBitMap(image_buffer + 62);
#endif
```

image

### DIFF
--- a/lib/gfxDraw/LICENSE
+++ b/lib/gfxDraw/LICENSE
@@ -1,0 +1,28 @@
+BSD 3-Clause License
+
+Copyright (c) 2024, Matthias Hertel
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+1. Redistributions of source code must retain the above copyright notice, this
+   list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright notice,
+   this list of conditions and the following disclaimer in the documentation
+   and/or other materials provided with the distribution.
+
+3. Neither the name of the copyright holder nor the names of its
+   contributors may be used to endorse or promote products derived from
+   this software without specific prior written permission.
+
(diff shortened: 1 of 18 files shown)


delhoume:
I have fixed some of the issues, but what seems to prevent the test from running is that none of BOARD_TRMNL or BOARD_WAVESHARE_ESP32_DRIVER is defined in the tests context.

delhoume:
an now it is because of ArduinoFake it seems. missing loop() and setup()
, so maybe it is not used at all, or there is a missing #define somewhere.

As you can see from no conflicts, my code changes are completely unrelated to what makes the tests fail, ad I am not at al familiar with the platform.io or frameworks used to test on

PR #114, opened 2025-04-16
More PNG tests

jrr:
Now that https://github.com/usetrmnl/firmware/pull/106 is in, time for the next step ;)

There are two versions of decodePNG(), one that takes buffers, and one that takes files.

This PR:
 - factors out their common bits into a new `processPNG()`
 - moves the buffer version into the library, since it's portable
   - (leaves the file version in the ESP project, since it isn't)
 - adds a handful of unit tests

These tests should be a good starting point for future PNG changes.


jrr:
@Fr3dr1ckson when you have a moment, this is rebased and passing again. It'd be nice to get this in, so that we could then add testcases covering the behavior from #116 .

PR #120, opened 2025-04-21
getting rid of unnecessary code

delhoume:
especially no flip necessary nor memory allocation to send the bitmap to the display.

same could be done for display_show_msg as we do not reuse the bitmap and this bitmap covers the whle screen.


delhoume:
Wire.h is included in Debug.h, i± do not see why, and it seems to mess with ArduinoFake in some way, even if it is not used at all

delhoume:
nevermind trying to merge these, too many changes at once.
In the future I will make much smaller PRs that can be quickly reviewed and merged and that will not break ci/cd in an unfixable way ;-)

PR #122, opened 2025-04-24
Add support for Waveshare E-Paper ESP32 Driver Board

skorokithakis:
I've added support to the official driver for the [Waveshare ESP32 driver board](https://www.waveshare.com/wiki/E-Paper_ESP32_Driver_Board), which makes it really easy to create a TRMNL with just hooking up an e-ink display to this driver board.

The changes required are minimal, as you can see, and I can add some #ifdefs for them. There should be one based on the environment (some of the changes are ESP32-specific), but there are things like faking the battery voltage for builds that don't use a battery.

Are there any specific defines you guys would like to see, or any opinions you have on this?

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -49,8 +49,8 @@ enum WIFI_CONNECT_RETRY_TIME // Time to sleep before trying to connect to the Wi
     WIFI_THIRD_RETRY = 300
 };
 
-#define PIN_RESET 9
-#define PIN_INTERRUPT 2
+#define PIN_RESET 25
+#define PIN_INTERRUPT 16
 #define PIN_BATTERY 3
 
 #define BUTTON_HOLD_TIME 5000
--- a/include/config.h
+++ b/include/config.h
@@ -49,8 +49,15 @@ enum WIFI_CONNECT_RETRY_TIME // Time to sleep before trying to connect to the Wi
     WIFI_THIRD_RETRY = 300
 };
 
+#if defined(BOARD_TRMNL)
+#define PIN_RESET 9
+#define PIN_INTERRUPT 2
+#elif defined(BOARD_WAVESHARE_ESP32_DRIVER)
 #define PIN_RESET 25
 #define PIN_INTERRUPT 16
+#define FAKE_BATTERY_VOLTAGE
+#endif
+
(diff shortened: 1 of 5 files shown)


skorokithakis:
I recommend enabling "hide whitespace" to review this, my editor didn't like the trailing whitespace everywhere and I stopped fighting its removal after a point.

skorokithakis:
This works fine with my board now (I don't have a TRMNL yet to check whether it works with the official device). The only backwards-incompatible change is that the default env is called `trmnl` now.

I've separated ESP32 from ESP32-C3 and added a `trmnl` env and a `waveshare-esp32-driver` env that use the above boards. This is for cleaner separation between the actual PCB (and thus the pins) and the architectures (as those have slight differences in the code).

ruohki:
Is the option for buttons and battery readouts included?

Fr3dr1ckson (approved):
Good job!

Tested on my og and waveshare board and got no issues
Liked what you've done with platformio.ini

It'll be good if you add button functions like on terminal (special functions or force refresh) or, if you managed to implement this functionality already  - provide steps for users to recreate

skorokithakis:
I'm afraid I don't have any buttons on my board to add that, but:

1) Those should just be straightforward pin reassignments, so the user can just specify the button pins, and
2) They'd be different for each user, so I'm not sure it makes much sense to include those in this PR.

raghavrithvik:
This build works to boot, but once wifi is connected, the device goes into a boot loop. Any idea where one needs to enter the BYOD mac address during firmware build?

preshit:
> This build works to boot, but once wifi is connected, the device goes into a boot loop. Any idea where one needs to enter the BYOD mac address during firmware build?

I've had the same issue.

skorokithakis:
Can you please post the logs from the UART? The MAC is entered on the Web interface.

ryanckulp:
> This build works to boot, but once wifi is connected, the device goes into a boot loop. Any idea where one needs to enter the BYOD mac address during firmware build?

you can input your BYOD mac in your device settings page. while logged into TRMNL > top-right dropdown > gear cog icon.

raghavrithvik:
Sorted. Thank you.

PR #125, opened 2025-04-28
support for variant V3 of BMP files / No Alpha value checking for palette 

delhoume:
rationale:
- do not reject files whose palette is not exactly what is expected for the unused Alpha value
- accept V3 BMP variant files as payload (they are 48130 instead of 48062 and are currently rejected)

68 bytes larger but same structure (also v4 with a few more bytes)

I had to change a bit the way payload size is checked to allow for slightly larger files
BMP parser now locates the colormap instead of harcoded  offset to 54
checking for reversed black and white does not look at Alpha value.

This should allow ./logo.bmp o be used as payload and previously rejected files to work

I added checked that new parsing works as before and new tests have been added for new method, based on logo.bmp


delhoume:
See https://en.wikipedia.org/wiki/BMP_file_format

added v4 variant and updated colormap location

PR #128, opened 2025-04-29
Enable WiFi Roaming feature

datacompboy:
This fixes late disconnect from distant access point in multi-AP (mesh) home networks; even if chip connects to weak-signal AP first, it get steered to better one nearby before the actual connection starts.


Omeryl:
don't we need `CONFIG_ESP_WIFI_SOFTAP_SUPPORT = n` to be y? does this break the captive portal?

ryanckulp:
another customer experiencing this issue provided this hint:
https://github.com/espressif/arduino-esp32/issues/2501

PR #130, opened 2025-05-07
Set HTTPS timeout to 15s instead of old 5s

datacompboy:
Set HTTPS timeout to 15s instead of old 5s to fight server-side slowness and weak Wifi signals.

### DIFF
--- a/src/bl.cpp
+++ b/src/bl.cpp
@@ -606,6 +606,8 @@ static https_request_err_e downloadAndShow()
   { // Add a scoping block for HTTPClient https to make sure it is destroyed before WiFiClientSecure *client is
 
     HTTPClient https;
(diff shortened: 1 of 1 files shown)


jamesthurley:
I created a similar PR a while ago, and in the end it was agreed to stick with the default timeout but make it configurable by the server:

https://github.com/usetrmnl/firmware/pull/87

I think the issues with long timeouts by default is battery drain and also the user ends up waiting a long time for an error screen if there is an issue (as it retries three times).

datacompboy:
When api/display is slow due to network issue or server overflow, there is no way to set timeout.
I agree the image download timeout may have sense to adjust, but just TCP retransmit can trivially add 3s to communication. With ssl, that means that 5s is only best case...

Issue #133, opened 2025-05-14
Add log levels


bkuhlmann:
## Why

Due seeing an increase in log levels since the release of Version 1.5.3, I'd like to suggest adding log levels to all `/api/log` requests (this also came up in team discussion during standup). Typically, this would be the following (starting with the most verbose to least verbose):

- **Debug**: Includes all log levels and is the most verbose.
- **Info**: Includes informational messages.
- **Warn**: Includes warning messages. Definitely worth being concerned about but should not break anything.
- **Error**: Includes error messages. These are bad not not catastrophic.
- **Fatal**: Includes fatal messages. These are catastrophic.

## How

Currently the log payloads look as follows (although, it would be nice to clean this up further as documented in this [issue](https://github.com/usetrmnl/firmware/issues/99)).

<details>
  <summary>Current Payload Example</summary>

ℹ️ The following uses Ruby hash syntax only for demonstration purposes.

``` ruby
{
  log: {
    logs_array: [
      {
        device_status_stamp: {
          battery_voltage: 4.772,
          current_fw_version: "1.2.3",
          free_heap_size: 160656,
          max_alloc_size: 990000,
          refresh_rate: 30,
          special_function: "none",
          time_since_last_sleep_start: 31,
          wakeup_reason: "timer",
          wifi_rssi_level: -54,
          wifi_status: "connected"
        },
        additional_info: {
          retry_attempt: 1
        },
        log_codeline: 597,
        log_id: 2,
        log_message: "Danger!",
        log_sourcefile: "src/bl.cpp",
        creation_timestamp: 1742000523
      }
    ]
  }
}
```
</details>

Given the above, this means we'd need to add a `log_level` key. Example (truncated for brevity and using Ruby syntax for demonstration purposes):

``` ruby
{
  log: {
    logs_array: [
        log_level: "info"
      }
    ]
  }
}
```

We'd need a way to update the device to provide different log levels as well so Core, and especially BYOS, could let the device be either more verbose or less. For BYOS, this would be powerful because we could start at log level INFO and when customers hit issues they could set log level to DEBUG to help file better bug reports.

Initially, this could be done by sending the desire log level as part of the API Display response to the device. Example (truncated for brevity):

``` json
{
  "log_level": "debug"
}
```

## Notes

This change would be of great help when debugging issues with customers both via Core but especially for BYOS.

jrr:
Today there are two kinds of logging:
```cpp
// send to the server else persist for later
[...]

PR #138, opened 2025-05-26
test stored_logs

jrr:
Gradually working toward #133, this PR moves `stored_logs.cpp` to the library and adds a few unit tests.

I wrapped the preferences access behind a new _Persistence_ interface, so that unit tests can use a mock in-memory alternative. For now it's only used for stored_logs, but in the future it could be used to help make other areas of the app testable.


src/bl.cpp
```diff
@@ -1852,7 +1853,8 @@ static void log_POST(char *log_buffer, size_t size)
   {
     Log_info("Was unable to send log to API; saving locally for later.");
     // log not send
-    store_log(log_buffer, size, preferences);
+    PreferencesPersistence preferencesPersistence(preferences);
```
jrr, line 1856: Starting small, for now `PreferencesPersistence` is used in just a few places. Down the road I imagine having a single instance replacing the app's use of `preferences`.

PR #150, opened 2025-06-03
Gfx font

delhoume:
This implements the Adafruit font format.

Current change are non-destructive, if merged then 
1) an official TRMNL font will have to be defined
2) .h can be generated with the fontconvert utility (part of Adafruit lib) - maybe in 4 sizes
3) Paint_DrawChar and Paint_DrawString_EN can be replaced by new functions
4) all WS Fontxx.cpp can be removed from repo. 

5) TODO: replace explicit font in drawing calls by a current font (as in Adafruit lib) so tha changing font does not need change in calling code.


delhoume:
Damn git mess too many changes

PR #151, opened 2025-06-03
gfxfont first impl

delhoume:
Tis just to check that it works for everybody.

If merged, we will have a few topics to consider
- define a corporate ttf to be used (or many, with sizes to precompute)
- create the .h
- replace all existing drawstrings
- remove fontxx.cpp
-  some other small changes (make current font instead of passing font  to drawing, add measuring functions)


delhoume:
obolete

PR #163, opened 2025-06-19
Add button support for Waveshare E-Paper ESP32 Driver Board

berezovskyi-oleksandr:
This PR extends the functionality introduced in #122 by implementing hardware button support for DIY units based on the Waveshare E-Paper ESP32 Driver Board using the RTC pin 33 as an interrupt source.

### What I found out

So turns out ESP32 doesn't like `gpio_wakeup_enable` for waking from deep sleep. Spent some time debugging this until I tried `esp_sleep_enable_ext1_wakeup` - works like a charm now.

The catch is that RTC wakeup needs the button pulled to GND, so I had to flip the logic in `read_button_presses()`. Not a big deal, just something to keep in mind.

### What works

Tested both long press and double click - everything's working smoothly. The unit wakes up from deep sleep reliably which was my main concern.

Let me know if you need me to test anything else or if you have questions about the implementation.

### DIFF
--- a/src/bl.cpp
+++ b/src/bl.cpp
@@ -32,6 +32,7 @@
 #include "driver/gpio.h"
 #include <nvs.h>
 #include <serialize_log.h>
+#include "driver/rtc_io.h"
 
 bool pref_clear = false;
 String new_filename = "";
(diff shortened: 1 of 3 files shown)


src/bl.cpp
```diff
@@ -1774,8 +1776,8 @@ static void goToSleep(void)
   esp_sleep_enable_timer_wakeup((uint64_t)time_to_sleep * SLEEP_uS_TO_S_FACTOR);
   // Configure GPIO pin for wakeup
 #if CONFIG_IDF_TARGET_ESP32
-  gpio_wakeup_enable((gpio_num_t)PIN_INTERRUPT, GPIO_INTR_LOW_LEVEL);
-  esp_sleep_enable_gpio_wakeup();
+  #define BUTTON_PIN_BITMASK(GPIO) (1ULL << GPIO)  // 2 ^ GPIO_NUMBER in hex
+  esp_sleep_enable_ext1_wakeup(BUTTON_PIN_BITMASK(PIN_INTERRUPT), ESP_EXT1_WAKEUP_ANY_HIGH);
```
Fr3dr1ckson, line 1780: Why cant we use ESP_EXT1_WAKEUP_ALL_LOW instead of ESP_EXT1_WAKEUP_ANY_HIGH so we wont invert pin in button.cpp

PR #170, opened 2025-06-27
version suffix for local/CI builds

jrr:
It's hard to keep track of whether my devices are on release builds or custom builds. This small change adds a suffix to tell them apart:

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -80,6 +80,8 @@ jobs:
         run: pip install --upgrade platformio
 
       - name: Build release firmware
+        env:
+          INCLUDE_GIT_HASH_IN_VERSION_NUMBER: true
         run: pio run -e trmnl
 
       - name: Name files
--- a/include/config.h
+++ b/include/config.h
@@ -5,10 +5,6 @@
 #define FW_MINOR_VERSION 5
 #define FW_PATCH_VERSION 9
 
-#ifndef FW_VERSION_SUFFIX
-#define FW_VERSION_SUFFIX ""
-#endif
-
 #define LOG_MAX_NOTES_NUMBER 5
 
 #define PREFERENCES_API_KEY "api_key"
--- a/include/config.h
+++ b/include/config.h
@@ -5,6 +5,10 @@
 #define FW_MINOR_VERSION 5
 #define FW_PATCH_VERSION 9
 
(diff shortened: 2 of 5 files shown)


bkuhlmann:
Ha, nice work!

By the way, the version metadata (i.e. commit) won't show up in the beta or production builds, correct? Only for local use?

jrr:
> the version metadata (i.e. commit) won't show up in the beta or production builds, correct? Only for local use?

That's the idea! It should only appear when using the "local" target (in platformio.ini). I'm not sure how the prod builds have been produced in the past[^1], but I expect them to use the "trmnl" target.  

[^1]: It'd be nice to bring some more rigor to the release process, and hopefully get most of it expressed in code in this repo. [Here's a start](https://github.com/usetrmnl/trmnl-firmware/pull/171).

bkuhlmann:
> That's the idea! 

Perfect.

> It'd be nice to bring some more rigor to the release process

I'd like that too because I see a few issues:

- The current [releases](https://github.com/usetrmnl/trmnl-firmware/releases/tag/v1.5.9) are hard to read and lacks a lot of detail in terms of understanding what the changes are.
- Would be nice to adhere to [Strict Semantic Versioning](https://alchemists.io/articles/strict_semantic_versioning). For example, 1.5.9 should have been 1.6.0 because a _new_ feature was introduced (i.e. minor bump, not a patch bump).
- Would be nice to use [Milestoner](https://alchemists.io/projects/milestoner) to automate the versioning and subsequent release note generation. ⚠️ Milestoner assumes a [Git Rebase](https://alchemists.io/articles/git_rebase) workflow, though. This is what [Terminus](https://github.com/usetrmnl/byos_hanami) uses if you need an example (I still need to host the HTML release notes and RSS feed, though).

Anyway, this is outside the scope of this code review but I can open a new issue to continue the discussion if you'd find that useful?

jrr:
Now I wired this up for CI builds, too. As of now:

- build target "local" -> `1.5.9+2cef89d`
- grab a recent build from CI -> `1.5.9+2cef89d`
  - (these are like "engineering builds", produced for each commit on `main`, using target "trmnl")
- build target "trmnl" manually -> `1.5.9`
  - (figuring the existing release process probably needs this)

Down the road I imagine another CI pipeline would produce official release builds from git tags, and those would also leave out the suffix.

version-suffix

PR #171, opened 2025-06-27
publish-build job

jrr:
Here's a new CI job which will produce (and upload!) a firmware image on every `main` commit. You can go in the job history and download it.

Eventually this could be hooked in to the [release](https://github.com/usetrmnl/trmnl-firmware/releases) process so that it doesn't depend on manually performing builds on anyone's workstation.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
 
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.11"
+          python-version: "3.x"
 
       - uses: maxim-lobanov/setup-xcode@v1
         if: matrix.os == 'macos-15'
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -55,3 +55,39 @@ jobs:
...
+      - uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cache/pip
+            ~/.platformio/.cache
+          key: ${{ runner.os }}-pio
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
```
beanieboi, line 77: Python 3.11 is pretty old at this point. Maybe is worth start with 3.13 here to save some future headaches :)

jrr: Good observation. I loosened it to 3.x to make this easier to maintain. (I often prefer to pin specific versions, but in this case I think it's unlikely that a future 3.x version breaks platformio)

PR #193, opened 2025-07-20
Combined log and submit

jrr:
This is the next (big) step toward the [plan](https://github.com/usetrmnl/trmnl-firmware/issues/133#issuecomment-2881605936) layed out in #133. Summary:

- `submit_log()` is removed
- instead, `Log_error` and `Log_fatal` also imply submitting to server
- instead of submitting immediately and falling back to local storage, logs are _stored_ immediately, and the stored logs are regularly flushed to the server. (see the calls to `submitStoredLogs()`)
  - this should result in fewer, but larger, requests to `/api/log`
  - and less risk of stack overflow when logging from within network interaction

Future:

- include the severity level in the submitted log (so that you'll see e.g. "error" and "fatal" on collected logs) (finally meeting the goal of #133)
- configurable submit threshold
- continue migrating e.g. `Log.info` calls to `Log_info`


bkuhlmann:
Looking forward to folding this in! A couple questions after reading through this since some of this isn't clear to me:

1. Will the JSON key be `level` in the payload? Example: `level: "error"`.
2. I noticed the use of `log_verbose` (case ignored for discussion). I assume that mean debug level?
3. Will the *warn* level be supported? I didn't see it in the changes but am curious.

PR #196, opened 2025-07-23
Disable light sleep in local development

jrr:
The new EPD code uses ESP32 light sleep (cool!). Unfortunately, this kills the serial connection and it doesn't seem to recover, so we lose a bunch of logs. It looks like this:

```
I: src/bl.cpp [238]: Display TRMNL logo start
I: src/display.cpp [242]: Paint_NewImage 0
I: src/display.cpp [243]: show image for array
I: src/display.cpp [279]: Display refresh start
Disconnected (read failed: [Errno 6] Device not configured)
Reconnecting to /dev/cu.usbmodem101 .......
```

This PR adds a condition to fall back to using `delay()` when working in the "local" environment. (e.g. `pio run -e local -t upload && pio device monitor -e local`)

Prod builds (and all other envs) will still get the light sleep.

I also added one more `submitStoredLogs()` call right before the new early WiFi disconnect, since the later one now won't be able to work.

### DIFF
--- a/lib/bb_epaper/src/bb_ep.inl
+++ b/lib/bb_epaper/src/bb_ep.inl
@@ -1673,8 +1673,12 @@ int bbepCreateVirtual(BBEPDISP *pBBEP, int iWidth, int iHeight, int iFlags)
 // Put the ESP32 into light sleep for N milliseconds
 void bbepLightSleep(uint32_t u32Millis)
 {
-  esp_sleep_enable_timer_wakeup(u32Millis * 1000L);
-  esp_light_sleep_start();
+#ifdef DO_NOT_LIGHT_SLEEP
+    delay(u32Millis);
+#else
+    esp_sleep_enable_timer_wakeup(u32Millis * 1000L);
+    esp_light_sleep_start();
+#endif
 }
 
 //
--- a/platformio.ini
+++ b/platformio.ini
@@ -129,6 +129,8 @@ build_flags =
 	-D ARDUINO_USB_CDC_ON_BOOT=1
 	-D WAIT_FOR_SERIAL=1
 	-D ARDUINOJSON_ENABLE_ARDUINO_STRING=1
+	# light sleep interrupts the serial connection and we miss logs
+	-D DO_NOT_LIGHT_SLEEP=1
 
 [env:seeed_xiao_esp32c3]
 extends = deps_app, esp32_base
--- a/src/bl.cpp
+++ b/src/bl.cpp
@@ -722,6 +722,8 @@ static https_request_err_e downloadAndShow()
             return HTTPS_WRONG_IMAGE_SIZE;
           }
 
+          submitStoredLogs();
+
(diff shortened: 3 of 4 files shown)


bitbank2:
I'm running on MacOS and I don't see this occur. The serial terminal continues to work after the light sleep. What is your exact setup?

jrr:
@bitbank2 Interesting! I wonder how our environments differ.

I'm on a Mac too, and primarily use the PlatformIO CLI (`pio`). Here are some versions:
```
pio --version
PlatformIO Core, version 6.1.18

# I don't imagine these matter, but:
uname -a
Darwin jrrmbp.local 24.5.0 Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:49 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6000 arm64
sw_vers
ProductName:		macOS
ProductVersion:		15.5
BuildVersion:		24F74
```

My typical development cycle is, with the device plugged in over usb-c:

- hold the button and flip the switch on
- `pio run -e local -t upload && pio device monitor -e local`
- (wait for it to finish flashing)
- flip the switch off, wait two seconds, flip it back on

I haven't tried using the PlatformIO IDE for this; perhaps it does something differently under the hood. What's your pattern?

Fr3dr1ckson:
I think the reason why serial is non working after light sleep written here: https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/api-guides/usb-serial-jtag-console.html

> When entering Light-sleep, the APB and USB PHY clock are gated. Thus, the USB Serial/JTAG controller is no longer able to receive or respond to any USB transactions from the connected host (including periodic CDC Data IN transactions). As a result:

> when entering Light-sleep, the USB Serial/JTAG device is unresponsive to the host/PC's USB CDC driver. The host/PC may then report the USB Serial/JTAG device as disconnected or erroneous (even if the USB cable is still physically connected).

> when exiting Light-sleep, it is possible that the host/PC does not re-enumerate (i.e., reconnect) the USB Serial/JTAG device given that the USB PHY's D+ line remains pulled up state during Light-sleep. Users may need to physically disconnect and then reconnect the USB cable.

So it seems that problem is within pc system configuration itself, so other users can struggle with this issue too. Since the only thing that users that dont want to remove light sleep from local is just erase DO_NOT_LIGHT_SLEEP flag inside local env, i think this pr is quite usefull and should be merged

bitbank2:
Up to this point, I have been keeping bb_epaper (the main branch) in sync with the changes for the TRMNL version, but this change would require that they diverge. It will cause future troubles for keeping both copies of the library in sync. Do you have a solution to the new problem it creates?

bitbank2:
The light sleep functionality is very specific to the ESP32; my library is generic and works on any target platform. I haven't released the light sleep version yet in the main branch. I suppose a macro as you propose would be fine to merge into the main repo.

PR #197, opened 2025-07-23
Cleanbmp

delhoume:
- make behaviour of BMPs similar to PNGs and new G5 images that are decoded only when needed for display
- do not reject some valid BMPs (and handle in displaying)
    - exact colormap is not checked in parseBMPHeader and negative height are accepted
  -  if black and white are inverted we reverse for display; if there are other colors (1bit bmps can have any 2 color palette) the we display the image with b&w colors, better than nothing imho
  -  bottom-up normal  layout is verticaly flipped when display only for positive height
 BMPs 
isPNG is stil there at the bl.cpp level but does not need to be passed to various display_show_xxx
and logo and qrcodes inernal bitmaps have already been converted to G5 so there are less pathes that we need to care about image being BMP

- fixed missing fillScreen ini display_show_msg, was showing for non WIFI_CONNECT messages.
image

- changed `png.cpp decodeBMP` to d`ecodeBMPFromMem` because there is a almost similar signature function in `png_file.cpp`

- updated BMP tests for new policy and added a top-down test mage


lib/trmnl/src/bmp.cpp
```diff
@@ -1,37 +1,38 @@
...
   uint32_t imageDataSize = *(uint32_t *)&data[34];
   uint32_t colorTableEntries = *(uint32_t *)&data[46];
-
-  if (width != 800 || height != 480 || bitsPerPixel != 1 || imageDataSize != 48000 || colorTableEntries != 2)
-    return BMP_BAD_SIZE;
-  // Get the offset of the pixel data
+ // Get the offset of the pixel data
   uint32_t dataOffset = *(uint32_t *)&data[10];
+// Display BMP information
+  printf("BMP Header Information:\r\nWidth: %d\r\nHeight: %d\r\nBits per Pixel: %d\r\nCompression Method: %d\r\nImage Data Size: %d\r\nColor Table Entries: %d\r\nData offset: %d", width, height, bitsPerPixel, compressionMethod, imageDataSize, colorTableEntries, dataOffset);
```
jrr, line 30: Why `printf` instead of `Log_info`?

lib/trmnl/include/png.h
```diff
@@ -15,4 +15,4 @@ enum image_err_e
 
 image_err_e processPNG(PNG *png, uint8_t *&decoded_buffer);
 
-image_err_e decodePNG(uint8_t *buffer, uint8_t *&decoded_buffer);
\ No newline at end of file
+image_err_e decodePNGFromMem(uint8_t *buffer, uint8_t *&decoded_buffer);
```
jrr, line 18: Fair. Those overloads were dangerously close 😳

src/display.cpp
jrr, line 1: It's really hard to review this with all the formatting changes. Can you discard al the non-meaningful changes?

(I'd like to get some standardized formatting rules in place, maybe clangformat or something, but we're not there yet)

lib/trmnl/include/bmp.h
```diff
@@ -9,4 +9,6 @@ enum bmp_err_e
   BMP_INVALID_OFFSET,
 };
 
-bmp_err_e parseBMPHeader(uint8_t *data, bool &reserved);
+#define BMP_SIGNATURE 0x4d42
```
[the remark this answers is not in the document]
delhoume: This one is inspired by Larry's code and allows to check for signature directly as 16 bit value on little endian machines 
image
 
image

.gitignore
```diff
@@ -8,3 +8,8 @@
 sdkconfig.*
 !sdkconfig.defaults
 
+._test.bmp
```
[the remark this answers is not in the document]
delhoume: I do not know where these fils come from.
maybe  a consequence of using  imageconvert

yes they need to be ignore as well

src/display.cpp
delhoume: should be much better, I committed a .clang-format in repo's root.
and manually made sure I have almost minimal changes in display.cpp

PR #201, opened 2025-07-27
Adds 2-bit grayscale, fast and partial updates to the original 7.5" eink panel.

bitbank2:
This change is relatively complex for the image handling. It was necessary to throw out the old PNG decode logic and start over. 2-bit (4gray) mode requires more memory than the ESP32-C3 has available, so I changed the PNG logic to decode and write to the EPD line by line. This doesn't hurt performance, but it does allow support for 4gray mode and partial updates.


src/display.cpp
```diff
@@ -521,18 +521,18 @@ void display_show_image(uint8_t *image_buffer, int data_size, uint8_t *image_buf
 #endif
         }
         bbep.writePlane(PLANE_0); // send image data to the EPD
-        iUpdateCount = 0; // reset the update count when woken by button press of BMP file is shown 
+//        iUpdateCount = 0; // reset the update count when woken by button press of BMP file is shown 
     }
     Log_info("Display refresh start");
 #ifdef BB_EPAPER
     if ((iUpdateCount & 7) == 0) {
```
ryanckulp, line 528: would this be safer as `iUpdateCount > 7`?

PR #204, opened 2025-07-30
Pr 201 fixes

ryanckulp:
via @bitbank2


ryanckulp:
16-shades--2bit

calendar--og-2bit

ryanckulp:
we're experiencing this issue: https://www.makerfabs.com/blog/post/cautions-in-using-esp32-adc-makerfabs-2

`analogReadMilliVolts()` response variance is not great, so we need a sufficient # of reads to calculate voltage. hard-coding 128 passes works, but adds latency and wastes power. as few as 8 passes is enough for some boards. playing with options to keep reading until summed `adc` "looks" real.

PR #216, opened 2025-08-10
fix for PNG payload > 48062 and malloc/free removed

delhoume:
Here are my propositions for the week-end\

- fix for PNG payload > 48062 reendered as static
- fix for G5 (small)logo processed as PNG (in loading screen)
- removal of most malloc/free  buffer in bl.cpp
- only clear screen when G5 bitmap is not fullscreen in display_show_msg x2)


bitbank2:
The code changes look good.

PR #218, opened 2025-08-12
withhttp for /api/setup and logo download

jrr:
Migrated the last couple uses of wificlient / httpclient to use the withHttp helper.

I've tested the happy path, but this could use some testing around the other scenarios. (unknown MAC, errors from /api/setup, etc.)

### DIFF
--- a/include/api-client/setup.h
+++ b/include/api-client/setup.h
@@ -1,0 +1,23 @@
+#pragma once
+
+#include <types.h>
+#include <api_types.h>
+#include <HTTPClient.h>
+
+struct ApiSetupInputs
+{
+  String baseUrl;
+  String macAddress;
+  String firmwareVersion;
+};
+
+struct ApiSetupResult
+{
+  https_request_err_e error;
+  ApiSetupResponse response;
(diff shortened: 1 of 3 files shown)


delhoume:
The wrong MAC address text comes from the server and is layed out by the horrible Paint_DrawMultilineText function.
This function is capable of word wrapping so the text comes unformatted from the server.
It was only handling fixed width fonts, @bitbank2 adapted it to work with the standard proportional TRMNL font but as you se eit is still not well rendered.

Larry has a replacement function that will position a paragraph with multiple lines (wityh explicit \n), centered horizontally whose code is 1 order of magnitude less than Paint_DrawMultilineText. I don't know if it handles word-wrapping

My advice is to use that function for every case in both display_show_msg, format the MAC sentence sent by the server with \n,and get rid of Paint_DrawMultilineText.

That willl make text placement much more simple and adaptable.

`display_centered_text` is in `multilang` branch but it should be easy to extract it and use it in main.
here is how clean it allows the display_show_msg case to be

image

PR #219, opened 2025-08-14
Include ID and Access-Token for Image

deisterhold:
Include ID and Access-Token when fetching image from server, but only if the image is hosted on the same host as the API.


deisterhold:
My thinking for including the access token is that so if the image contains sensitive data (financial, calendar, etc.) that some form of auth should be required. The reason for only including if it's on the same host is in the case that the alias plugin is used, then the token isn't shared with a third party.

PR #238, opened 2025-09-29
Breakpoint typo

dvankley:
Looks like this is a typo, since there is no symbol `setupy` and `setup` is a typical configuration.


ryanckulp:
thanks @dvankley !

Issue #244, opened 2025-11-07
HTTPClient doesn't properly handle compressed response without Content-Length header when serving images through proxy/cache (Cloudflare)


tajchert:
This is awesome project and I'm using it as BYOD/S on XIAO-ESP32-S3 with [byos_hanami](https://github.com/usetrmnl/byos_hanami) self hosted server and exposed via Cloudflare.
In such case (exposed via Cloudflare) response image is returned without some headers (crucially without `Content-Length`) and with gzip compression due to its cache/CDN.

In such case device cannot properly parse response and is stuck rebooting, logs from device (firmware: 1.6.10):
```
I: lib/trmnl/include/http_client.h [25]: ==== withHttp() https://trmnl.mtajchert.com/uploads/d3ea2d3cc3d06ae949d0edefbecf3a22.png
I: src/bl.cpp [675]: GET...
I: src/bl.cpp [676]: RSSI: -60
E: src/bl.cpp [702]: [HTTPS] GET... code: 200
I: src/bl.cpp [703]: RSSI: -55
I: src/bl.cpp [711]: Content size: -1
I: src/bl.cpp [722]: RSSI: -55
I: src/bl.cpp [723]: Stream timeout: 1000
I: src/bl.cpp [725]: Stream available (may show as 0 and that's okay): 728
I: src/bl.cpp [729]: Starting a download at: 1762551184
Guru Meditation Error: Core  1 panic'ed (StoreProhibited). Exception was unhandled.

Core  1 register dump:
PC      : 0x42033117  PS      : 0x00060330  A0      : 0x820d27bc  A1      : 0x3fceba80  
  #0  0x42033117 in Stream::readBytes(char*, unsigned int) at /Users/mtajchert/.platformio/packages/framework-arduinoespressif32/cores/esp32/Stream.cpp:289

A2      : 0x00000000  A3      : 0x00000000  A4      : 0xffffffff  A5      : 0x3fcaab04  
A6      : 0x00000010  A7      : 0x00000064  A8      : 0x82033114  A9      : 0x3fceba60  
A10     : 0x00000036  A11     : 0x00000000  A12     : 0x00000024  A13     : 0x00000000  
A14     : 0x00000002  A15     : 0x00000064  SAR     : 0x00000010  EXCCAUSE: 0x0000001d  
EXCVADDR: 0x00000000  LBEG    : 0x400570a4  LEND    : 0x400570a9  LCOUNT  : 0x00000000  


[...]

PR #245, opened 2025-11-07
Fix for image served with a proxy/CDN - without Content-Length header and with a compression

tajchert:
This resolves issue of #244 (Guru Meditation StoreProhibited) when using BYOD/S with a proxy/cache that modifies image response from a web server for TRMNL.

**Issues fixed:**
- Null pointer dereference when malloc(-1) failed due to missing Content-Length
- Chunked transfer encoding markers corrupting image data
- Server gzip compression preventing proper image decoding

**Changes:**
- Use HTTPClient getString() to handle chunked encoding automatically
- Add Accept-Encoding: identity header to request uncompressed data
- Add malloc NULL check with HTTPS_OUT_OF_MEMORY error code
- Improve downloadStream with disconnect detection and 5s idle timeout

### Description
**Root Cause:** Servers using chunked transfer encoding without Content-Length headers caused malloc(-1) to return NULL, leading to a crash when writing to the buffer. Additionally, chunked encoding markers were being read as image data, corrupting the PNG files.

**Solution:** Replaced raw stream reading with HTTPClient::getString() which automatically decodes chunked transfers. Added Accept-Encoding: identity header to prevent gzip compression. Implemented malloc NULL checking with new HTTPS_OUT_OF_MEMORY error code.

**Additional Improvements:** Enhanced downloadStream() with disconnect detection and 5-second idle timeout for better reliability with other download paths in the codebase.

This code was created using Claude Code, however I put effort to check, document and test it.


ryanckulp:
hey @tajchert, can you email me? trying to hook you up with a [Creator Fund](https://usetrmnl.com/blog/creator-fund) earnout. ;) ryan@usetrmnl.com

Issue #270, opened 2026-01-03
Waveshare 7.5 Inch display renders the screens more ghosted instead of clear.


SBajonczak:
Hi I build the TRMNL Firmware via PIO for Waveshare displays I own the three color 7.5 inch display (v3). 
So actually I got this behavior. While it boots or changing display it will display this with a clear icon 

<img width="889" height="570" alt="Image" src="https://github.com/user-attachments/assets/b68de05d-af6d-4f89-baf6-f2296426bbdc" />

But when It tries to display It will render it as ... yeah ..ghosted image like this: 

<img width="986" height="578" alt="Image" src="https://github.com/user-attachments/assets/e060141a-e30a-4e6c-8a99-e0e42f6c0009" />

There is not a display issue because I tests it with another firmware with ESPHome. So I think there is a small problem in my config or similar. 
It there any known issue about this? 

Here is a log output of the screen. 

For Information I use a self hosted system. I frun a laravel webhost and the device is connected to an Waveshare hat via an esp from waveshare. 

```
I: src/display.cpp [81]: dev module end
I: src/filesystem.cpp [21]: SPIFFS mounted
I: src/bl.cpp [252]: Display TRMNL logo start
I: src/display.cpp [915]: display_show_image start
I: src/display.cpp [916]: maximum_compatibility = 1

I: src/display.cpp [974]: Display refresh start
I: src/display.cpp [982]: src/display.cpp [982]: Forcing full refresh; desired refresh mode was: 2    

I: src/display.cpp [992]: src/display.cpp [992]: EPD refresh mode: 2

I: src/display.cpp [1002]: display_show_image end
I: src/bl.cpp [258]: Display TRMNL logo end
I: src/bl.cpp [262]: Firmware version 1.7.2
I: src/bl.cpp [263]: Arduino version 2.0.17
I: src/bl.cpp [264]: ESP-IDF version 4.4.7
I: src/filesystem.cpp [206]: Filesystem Usage: 6024/233681
I: src/filesystem.cpp [211]:   2574  last.png
I: src/filesystem.cpp [211]:   2751  current.png
I: src/bl.cpp [2287]: NVS Usage: 271/630 entries (43.02
I (1764) wifi:wifi driver task: 3ffc630c, prio:23, stack:6656, core=0
I (4251) wifi:wifi firmware version: 1fd20f4
I (4251) wifi:wifi certification version: v7.0
I (4251) wifi:config NVS flash: enabled
[...]

PR #272, opened 2026-01-09
Added support for Waveshare 3color 7.5inch display

SBajonczak:
As I mentioned I got hte problem with ghosting and scanlines in using the waveshare 7.5inch Display. 

I figured out that theacutal implementation is only suitable for the B/W variant and not for the three colors variant. 

I added the missing start sequence in this PR to support the three colors waveshare display. 

Referencing this issue for the initial bug: https://github.com/usetrmnl/trmnl-firmware/issues/270

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,4 +1,6 @@
 {
+    // See http://go.microsoft.com/fwlink/?LinkId=827846
+    // for the documentation about the extensions.json format
     "recommendations": [
         "pioarduino.pioarduino-ide",
         "platformio.platformio-ide"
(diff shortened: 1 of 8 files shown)


bitbank2:
Sorry, I'm not sure what you mean. The B/W and B/W/R updates are exactly the same (at least in my code). The display only has 1 update mode built in. I will be adding a new configuration to trmnl-firmware for the Seeed Studio epaper PCB (ESP32-S3) connected to the B/W/R 7.5" panel. This requires a lot of new code in trmnl-firmware to match color images to B/W/R. Displaying 1-bit B/W content on the B/W/R panel is not much different from the current code.

PR #296, opened 2026-01-29
Add Ed25519 challenge-response authentication

oetiker:
## Summary

Adds Ed25519 challenge-response authentication as an alternative to API key auth, giving each device a unique cryptographic identity.

### Why this matters

API keys are shared secrets — if one is intercepted, an attacker can impersonate the device indefinitely. Ed25519 authentication eliminates this class of attack:

- **No shared secrets on the wire.** The private key never leaves the device. Only the public key and a timestamped signature are transmitted.
- **Replay-resistant.** Each request is signed with a server-provided timestamp, so captured requests cannot be reused outside the validity window.
- **Tamper-evident identity.** The keypair is generated on-device and stored in NVS. A device's public key becomes its unforgeable identity — the server can detect if a device has been cloned or factory-reset.
- **One additional round-trip.** The device fetches the server timestamp via `/api/time` before signing. This ensures clock-independent replay protection.
- **Server-controlled rollout.** The server sets `auth_mode` per device via `/api/setup`, so existing devices continue working unchanged.

### What changed

- Ed25519 keypair generation and NVS persistence (`device_identity.cpp`)
- Challenge-response signing: `sign(timestamp_ms || public_key)` (`auth_signature.cpp`)
- TweetNaCl Ed25519 implementation with hardening: signature malleability rejection (S < L check), VLA stack guard
- Key generation moved after WiFi connect for RF-based `esp_random()` entropy
- API clients send `X-Public-Key`, `X-Signature`, `X-Timestamp` headers when in ed25519 mode
- Shared `bytesToHex` extracted to `hex_utils.h`

## Test plan
- [ ] `pio run` builds cleanly
- [ ] Device generates and persists keypair on first boot
- [ ] `X-Public-Key`, `X-Signature`, `X-Timestamp` headers sent when `auth_mode: "ed25519"`
- [ ] Factory reset regenerates keypair
- [ ] Fallback to API key auth when server returns `auth_mode: "api_key"`


oetiker:
The point of this PR is to prevent easy cloning of a trmnl device with this feature only reading the nv private key in the device allows cloning it. This feature is important when one wants to be sure to only display certain information on a particular device. In the current setup, only the api key is known, the information destined for a particular device can be downloaded freely.

NiKiZe:
Current implementation does no verification of https connections, which means if you control the network, you can mitm and grab apikeys and data, just so there is no illusions of security.

oetiker:
@NiKiZe yes, this feature does not prevent MM on the actual information transmissions, but it does stop a third-party from also downloading the information independently.

schrockwell:
@oetiker, thanks for the contribution but we are going to close this for now.

We have a couple solutions for keeping TRMNL access secure. Today we have BYOS where you can run on a trusted network that won't run afoul of attacks, and we are working on TLS client certificate support as the solution for end-to-end security. It's a simpler and more widely-adopted approach to solving this problem.

oetiker:
mtls is way more complex than using an ed25519 signature ... I am in the process of developing a solution where it is of primary importance to have an easy way making display devices unique ...

oetiker:
brought it up to date with current main

PR #299, opened 2026-01-30
Add Board, Colors, and Measured-Colors headers to display API requests

oetiker:
## Summary
- Define per-board `DEVICE_BOARD`, `DISPLAY_COLORS`, and `DISPLAY_MEASURED_COLORS` macros in `config.h` for all supported hardware variants
- Add `displayColors`, `displayMeasuredColors`, and `board` fields to `ApiDisplayInputs`
- Send `Colors`, `Board`, and `Measured-Colors` HTTP headers in display API requests

## Why
The backend currently has no way to know which board variant or display color palette a device supports. By sending these as headers, the API can return optimized images matched to the actual hardware — e.g. serving 4-color images to `trmnl_og_4clr` or 16-shade grayscale to `trmnl_x`.

`Measured-Colors` provides the actual sRGB values the e-ink panel produces (derived from Pimoroni's photographed saturated palette and community spectrophotometer data), enabling better server-side dithering and quantization compared to idealized palette values.

## Test plan
- [ ] Verify all three headers appear in API requests on physical hardware or via network capture
- [ ] Confirm correct values for each board variant build


ryanckulp:
thanks for this. i'm curious why it's needed however. currently our Device Model picker maps the hardware's capabilities (colors, selected palette, screen size, etc) to our render pipeline, which generates images according to those specs and then returns that graphic to the device.

https://trmnl.com/api/models
https://trmnl.com/api/palettes

i could see this being useful if (for example) a Kindle and a Nook were requesting content with the same API key, as it would allow the `/api/display` endpoint to shuffle around the available (already generated) Screens to return the one being requested. but that is not officially supported. 

user sets their device's model + palette preferences, and all requests for content with that device's API key incorporate those specs.

schrockwell:
I agree with @ryanckulp - this is duplicate information that is already provided via the TRMNL API, or for DIY devices the user could specify it out-of-band. The `Model` header should be enough for servers to uniquely identify the model and  palette. Arguably, we should be _removing_ headers like `Width` and `Height`.

It also does not account for displays which support large color palettes, like 16-bit or 32-bit color.

In general, I think this approach is a slippery slope. It will lead to out-of-sync palettes between firmware and server implementations. At the end of the day, it's the server's responsibility to render an appropriate image.

Issue #266, opened 2025-12-17
Issues Retrieving Larger Images with FW1.7.1


kylem0:
This commit seems to be causing problems on my trmnl device: https://github.com/usetrmnl/trmnl-firmware/commit/2944f73ebdf14ecc1011a99752c21b5b0cc7fdc5

I have my device pointed to a simple server I wrote myself, which could be the cause of my problems, but FW 1.6.9-1.7.0 don't cause any problems for me.

The device is sending this log to the server occasionally when trying to retrieve some image files from my server:

`{"logs":[{"created_at":1765951040,"id":103,"message":"Receiving failed. No data received","source_line":740,"source_path":"src/bl.cpp","wifi_signal":-54,"wifi_status":"connected","refresh_rate":1800,"sleep_duration":85,"firmware_version":"1.7.1","special_function":"none","battery_voltage":3.998,"wake_reason":"timer","free_heap_size":171412,"max_alloc_size":147456,"retry":1}]}`

That error was added/changed in this commit. I am a novice developer and don't know much about writing code for the ESP32, but my best guess right now is related to memory. The buffer that was used before this commit seemed to work fine with the same image. The content-length of the image from the log above is 65932, I'm displaying 2-bit PNG images.

Reading [this doc for http.getString()](https://avantmaker.com/references/esp32-arduino-core-index/esp32-httpclient-library/esp32-httpclient-library-getstring/) it mentions getString is better for small to medium payloads. Could the buffer code used previous be better for retrieving larger images? Or are my images just too large?

For now I reverted to 1.7.0 and things appear to be working fine.

kylem0:
Updated to 1.7.6 and tried a few random images and I no longer see and problems retrieving larger images. I think this can be marked resolved. Thank you for the fix!

PR #301, opened 2026-02-02
scoped getString

delhoume:
fix for https://github.com/usetrmnl/trmnl-firmware/issues/266
using getString consumes much more memory than previous code so while double allocation of payload size is "fixed", max payod sze  is reduxcesd from 90000 to maybe 65000.

Hopefully max png size for 800x480x2bit should stay under this limit, my current largest is 61274


ryanckulp:
thanks again for the PR, implemented #310 which gets us all the way to 95kb.

PR #304, opened 2026-02-06
Add E1001 (Seeed reTerminal) display support

oetiker:
## Summary

- Adds display support for **BOARD_SEEED_RETERMINAL_E1001** (GDEY075T7 7.5" 800x480 panel, UC8179 controller)
- Custom UC8179 init sequence matching [Seeed_GFX reference](https://github.com/Seeed-Studio/Seeed_GFX/blob/master/TFT_Drivers/UC8179_Init.h), needed because bb_epaper doesn't natively support this panel yet
- 4-gray rendering with correct UC8179 buffer order (OLD/NEW) and tuned booster settings
- SD card SPI bus contention fix (GPIO16) shared with E1002
- `PLANE_DUPLICATE` for 1-bit mode (UC8179 needs both OLD and NEW buffers written)

Only `src/display.cpp` is modified. All changes are behind `#ifdef BOARD_SEEED_RETERMINAL_E1001` guards so no other board variants are affected.

## Test plan

- [x] Builds cleanly: `pio run -e seeed_reTerminal_E1001`
- [x] Tested 1-bit (B/W) rendering on physical E1001 hardware
- [x] Tested 4-gray rendering on physical E1001 hardware
- [ ] Verify no regression on other board targets (build-only)

### DIFF
--- a/src/display.cpp
+++ b/src/display.cpp
@@ -29,6 +29,12 @@ BBEPAPER bbep(EP75R_800x480);
     {EP75YR_800x480, EP75YR_800x480}, // b = darker grays
 };
 BBEPAPER bbep(EP75YR_800x480);
+#elif defined(BOARD_SEEED_RETERMINAL_E1001)
+    {EP75_800x480_GEN2, EP75_800x480_4GRAY_GEN2}, // default (for GEN2 panel)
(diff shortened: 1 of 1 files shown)


Fr3dr1ckson:
Closed due to already added support for this device

PR #315, opened 2026-02-24
Spiffs optimization

bitbank2:
The purpose of this PR is to improve battery life, reduce AWS server bandwidth and reduce FLASH storage wear. It accomplishes all of this by saving plugin images on the ESP32's FLASH file system and avoiding image downloads when not needed. Here are the detailed changes to the behavior:

BEFORE:
The current image filename is compared with the previous, if it's the same then the download and image updates are skipped. If the image name is different, the old image is renamed from current.png to previous.png (in FLASH), the new image is written to FLASH as current.png and the full plugin filename is written to preferences (since it can be longer than 31 characters). This logic will only optimize the rare case of a user having one item in their playlist. The average user now has 5+ items in their playlist.

AFTER:
The current image filename is compared to all of the images stored in FLASH. If it exists, the download is skipped and the image from FLASH is displayed. If the filename is not found, it's stored in FLASH. The filename is adjusted to fit within the 31 character limit of SPIFFS (retaining its unique ID and timestamp). While searching for a matching name in FLASH, older versions of the same plugin/mashup are deleted as well as images older than 24h. The purge of old files ensures that the FLASH doesn't fill up if the user changes their playlist.

BENEFITS:
For the average user with a reasonable number of plugins that don't change frequently, significant AWS bandwidth and FLASH erasure is avoided. Significant battery life is also saved by avoiding the AWS download and the erasure of FLASH. Here's a new vs old example day (user has 4 plugins - days remaining, weather, new movies, hacker news):

OLD - 
Every playlist event throughout the day will cause an AWS download and FLASH erasure/writing.

NEW -
The first time each plugin is displayed, it will cause an AWS download and write to FLASH. Subsequent playlist events will skip the download and FLASH erasure/writing for the majority of refreshes. Weather and Hacker News might update every few hours, but the slower changing plugins will be able to skip downloading throughout the day.

The FLASH partition of the OG is defined to be 0x40000 (256K bytes). This is enough to store 10-20 average plugin images. Enough to cover the vast majority of TRMNL users who have 5+ plugins active.

### DIFF
--- a/src/bl.cpp
+++ b/src/bl.cpp
@@ -2116,7 +2116,7 @@ static void writeSpecialFunction(SPECIAL_FUNCTION function)
     Log.info("%s [%d]: SF saved. Reading...\r\n", __FILE__, __LINE__);
     if ((SPECIAL_FUNCTION)preferences.getUInt(PREFERENCES_SF_KEY, 0) == function)
     {
-      Log.info("%s [%d]: No needed to re-write\r\n", __FILE__, __LINE__);
+      Log.info("%s [%d]: No need to re-write\r\n", __FILE__, __LINE__);
     }
     else
     {
--- a/src/display.cpp
+++ b/src/display.cpp
@@ -148,6 +148,7 @@ void display_reset(void)
 {
     Log_info("e-Paper Clear start");
     bbep.fillScreen(BBEP_WHITE);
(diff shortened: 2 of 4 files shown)


Fr3dr1ckson:
Tnx for PR, looking forward to test it

PR #326, opened 2026-03-25
Add support for Inkplate 10

craftbyte:
This adds support for the Inkplate 10.

Currently it uses my fork of FastEPD, until the upstream gets merged. See https://github.com/bitbank2/FastEPD/pull/28 for status.


lanrat:
Just a heads up, I maintain a different firmware for the Inkplate set of devices with full Trmnl support

https://github.com/lanrat/homeplate

schrockwell:
@bitbank2 Do we already support Inkplate 10 under a PlatformIO environment with a different name? We have supported it as BYOD in the Core web app for a long time, so I was surprised not to find any "inkplate" references in the firmware.

ryanckulp:
i think we hit pause on native support as Homeplate maintains a fork:

https://github.com/lanrat/homeplate

schrockwell:
Thanks @ryanckulp. We are going to close this based on the existing support for Inkplate 10 in Homeplate. Thank you for the work @craftbyte.

PR #335, opened 2026-04-21
feat: add Xteink X3 board support (BOARD_XTEINK_X3)

sermelipharo:
Hi! Following up on the suggestion to mirror the X4 pattern — here's X3 support in the same style.

## What changed

| File | Change |
|---|---|
| `platformio.ini` | New `[env:xteink_x3]` env |
| `src/DEV_Config.h` | X3 SPI pin block |
| `include/config.h` | `DEVICE_MODEL`, `PIN_INTERRUPT`, `PIN_BATTERY`, correct BMP size for 792×528 |
| `src/display.cpp` | `EP368_792x528` / `EP368_792x528_4GRAY` panel selection |
| `src/bl.cpp` | Extend GPIO13 battery-latch hold to cover X3 |

## Hardware

The Xteink X3 is an ESP32-C3 e-reader with a 792×528 SSD1677-family panel. The SPI pin layout is identical to the X4 so `DEV_Config.h` is a straight copy. The non-obvious bit is the battery latch: the X3 (like the X4) uses a MOSFET-off topology where GPIO13 must be held HIGH through deep sleep or the device cuts its own power and can never timer-wake. The `bl.cpp` change just adds `BOARD_XTEINK_X3` to the existing `#if defined(BOARD_XTEINK_X4)` block.

`EP368_792x528` / `EP368_792x528_4GRAY` are already present in bb_epaper — no library changes needed.

## Not included yet

Battery readout via the BQ27220 I²C fuel gauge (the X3 has a proper fuel gauge instead of a bare ADC divider). `PIN_BATTERY 0` gives a raw ADC read that works but isn't accurate. Happy to add BQ27220 support in a follow-up if that's in scope.

## Test plan

- [x] Builds cleanly with `pio run -e xteink_x3`
- [x] Flashed on real X3 hardware; device provisions, fetches display, and timer-wakes correctly
- [ ] 4-gray / partial refresh not yet exercised — would appreciate testing if you have time

### DIFF
--- a/.github/workflows/publish-firmware.yml
+++ b/.github/workflows/publish-firmware.yml
@@ -59,7 +59,7 @@ jobs:
         run: cp .pio/build/trmnl/firmware.bin ${{ env.APP_FIRMWARE_FILE }}
 
       - name: Upload firmware artifacts
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: firmware images
           path: "*.bin"
--- a/include/config.h
+++ b/include/config.h
@@ -2,8 +2,8 @@
 #define CONFIG_H
 
 #define FW_MAJOR_VERSION 1
-#define FW_MINOR_VERSION 7
-#define FW_PATCH_VERSION 8
+#define FW_MINOR_VERSION 8
+#define FW_PATCH_VERSION 0
 
 // Helper macros for stringification
 #define STRINGIFY(x) #x
--- a/include/config.h
+++ b/include/config.h
@@ -37,8 +37,13 @@
 
 #define WIFI_CONNECTION_RSSI (-100)
 
+#ifdef BOARD_XTEINK_X3
+#define DISPLAY_BMP_IMAGE_SIZE 52334 // 62-byte header + 52272 bytes bitmap (792*528 1bpp / 8)
+#define DEFAULT_IMAGE_SIZE 52272
+#else
 #define DISPLAY_BMP_IMAGE_SIZE 48062 // in bytes - 62 bytes - header; 48000 bytes - bitmap (480*800 1bpp) / 8
 #define DEFAULT_IMAGE_SIZE 48000
+#endif
 #ifdef BOARD_TRMNL_X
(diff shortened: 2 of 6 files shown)


bitbank2:
Thanks for the effort, but we've already got this covered. I'll merge my new code after some higher priority issues have been resolved.

@Fr3dr1ckson please don't merge/work on this.

@sermelipharo the addition of the X3 requires a bit more effort than you/claude have done - since we don't have dynamic image size in the rendering engine, the 800x480 images would leave an ugly gap of 48 rows with random contents. I wrote a fix for that.

sermelipharo:
Thanks a lot for the quick reply and the direct feedback 🙏 My bad for opening this cold instead of pinging you first to ask about the X3 roadmap. Closing this PR now, better keep your flow, because I'm not so much experienced with this product.

If it's useful to anyone hitting this in the meantime, I've been running a small fork at [sermelipharo/x3-trmnl](https://github.com/sermelipharo/x3-trmnl), smoke-tested end-to-end against LaraPaper and a BYOS Next.js instance on real hardware (captive-portal provisioning, `/api/setup`, `/api/display`, `reset_firmware`, OTA via `update_firmware`, BMP + PNG decode, timer-wake on battery, nested-L calibration confirming 792×528 with ~1–2 px bezel). More than happy to fold anything from there into mainline when your version is ready — just ping and I'll redo it against your branch.
 
On the 48-row noise concern, that's totally a real problem for an engine that doesn't pre-clear, and the proper fix is definitely dynamic sizing like you're doing. In the fork I'm sidestepping it the cheap way (`display.clearScreen(0xFF)` before every decode in `src/trmnl/TrmnlImage.cpp`, so the uncovered band ends up white instead of random), which is a different codebase than yours but might be a useful one-liner as a safety net alongside the real fix. Sharing just in case, zero expectation.

Thanks again! I'm genuinely excited to see what you land

ritiek:
> Thanks for the effort, but we've already got this covered. I'll merge my new code after some higher priority issues have been resolved.

Hi @bitbank2, I've been looking forward to X3 support as well. Is there a branch I can test against? No rush, happy to wait.

balaclava-guy:
@bitbank2 is there a PR we can follow for X3 support?

PR #337, opened 2026-04-23
Ethernet-over-USB support with WiFi fallback

cerebrumdk:
Adds CDC-ECM Ethernet (RTL8152/RTL8153 USB dongle) as the primary network interface for TRMNL_X, with automatic fallback to WiFi when no dongle is present.

## New library: lib/ethernet/

connection_manager — FreeRTOS event-group-based interface tracker:
  - conn_mgr_init() / conn_mgr_mark_ready() / conn_mgr_mark_lost()
  - conn_mgr_is_eth() / conn_mgr_is_wifi() / conn_mgr_is_ready()
  - conn_mgr_wait_any(ticks) — blocks until any interface is up
  - conn_mgr_update_default_route() — promotes Ethernet over WiFi as the default netif when both are present
  - All getters return safe defaults before conn_mgr_init() is called (handles early wakeup paths that run before network init)

ethernet_config — CDC-ECM driver bootstrap:
  - ethernet_start() initializes the USB host driver for RTL8152/8153
  - netif event handler calls conn_mgr_mark_ready/lost on DHCP acquire/loss and triggers default route update

## New components (moved from managed_components/ to components/)

cdc_ecm_host library by default is downloaded to managed_components/ via idf file, but it's gitignored and wiped on idf.py update-dependencies. Since I've made changes to this lib, I need it to be persistent, hence moved to components. I also added function to espressif__usb espressif lib, hence needed to move it to components for the same reason, so they survive dependency updates.

cdc_ecm_host:
  - Sends SET_CONFIGURATION(2) to switch RTL8152 from vendor to CDC-ECM mode at enumeration time
  - Calls usbh_dev_update_config_desc() to refresh USBH's internal descriptor cache after the config switch
  - Verbose USB descriptor dump logs commented out

espressif__usb:
  - Added usbh_dev_update_config_desc() — same as usbh_dev_set_config_desc() but without the ADDRESS-state requirement, needed because RTL8152 is already CONFIGURED when we switch configurations

## Boot flow (BOARD_TRMNL_X)

1. esp_netif_init + event loop + conn_mgr_init + ethernet_start
2. Poll up to 5s for ETH_READY_BIT (breaks immediately on DHCP)
3. If no Ethernet after 5s: modem init -> WiFi autoConnect -> captive portal
4. WiFi connect/disconnect hooks into conn_mgr so default route and network-availability checks stay accurate throughout the cycle

## Other changes

- For getting MAC address get_device_mac() is used now instead of WiFi.macAddress() in case user doesn't connect to WiFi from the beginning
- getDeviceStatusStamp() reports ethernet / wifi / none with appropriate RSSI (0 for ethernet)
- WifiCaptivePortal.resetSettings() guarded by conn_mgr_is_wifi()


ryanckulp:
hey @cerebrumdk, would you mind rebasing this and opening a new PR into `main`, we had a massive X <> main merge recently. apologies for not clarifying this request as i initially closed yours.

cerebrumdk:
Hi @ryanckulp ! Sorry for the late response, opened a new PR: https://github.com/usetrmnl/trmnl-firmware/pull/361

PR #348, opened 2026-05-08
Add printf attribute to logging functions and fix any warnings

PGIII:
Adds a macro that enables the printf attribute on GCC/Clang.
It seems that the ESP IDF adds -Wno-format.
So this enables -Wformat for just the exe (in CMake) and just the trmnl lib (in PlatformIO).
I also fixed any warnings that came up.

### DIFF
--- a/lib/trmnl/include/compiler_attrs.h
+++ b/lib/trmnl/include/compiler_attrs.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <inttypes.h>
+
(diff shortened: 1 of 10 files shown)


schrockwell:
Thank you! Any compile-time checks we can add to prevent runtime crashes is a win in my book.

`main` had diverged quite a bit, so I rebased this branch onto `main` and re-fixed the current batch of format warnings.

PGIII:
Thanks for merging, happy to help!

PR #361, opened 2026-05-20
Ethernet-over-USB support with WiFi fallback

cerebrumdk:
Adds CDC-ECM Ethernet (RTL8152/RTL8153 USB dongle) as the primary network interface for TRMNL_X, with automatic fallback to WiFi when no dongle is present.

connection_manager — FreeRTOS event-group-based interface tracker:
  - conn_mgr_init() / conn_mgr_mark_ready() / conn_mgr_mark_lost()
  - conn_mgr_is_eth() / conn_mgr_is_wifi() / conn_mgr_is_ready()
  - conn_mgr_wait_any(ticks) — blocks until any interface is up
  - conn_mgr_update_default_route() — promotes Ethernet over WiFi as the default netif when both are present
  - All getters return safe defaults before conn_mgr_init() is called (handles early wakeup paths that run before network init)

ethernet_config — CDC-ECM driver bootstrap:
  - ethernet_start() initializes the USB host driver for RTL8152/8153
  - netif event handler calls conn_mgr_mark_ready/lost on DHCP acquire/loss and triggers default route update

cdc_ecm_host library by default is downloaded to managed_components/ via idf file, but it's gitignored and wiped on idf.py update-dependencies. Since I've made changes to this lib, I need it to be persistent, hence moved to components. I also added function to espressif__usb espressif lib, hence needed to move it to components for the same reason, so they survive dependency updates.

cdc_ecm_host:
  - Sends SET_CONFIGURATION(2) to switch RTL8152 from vendor to CDC-ECM mode at enumeration time
  - Calls usbh_dev_update_config_desc() to refresh USBH's internal descriptor cache after the config switch
  - Verbose USB descriptor dump logs commented out

espressif__usb:
  - Added usbh_dev_update_config_desc() — same as usbh_dev_set_config_desc() but without the ADDRESS-state requirement, needed because RTL8152 is already CONFIGURED when we switch configurations

1. esp_netif_init + event loop + conn_mgr_init + ethernet_start
2. Poll up to 5s for ETH_READY_BIT (breaks immediately on DHCP)
3. If no Ethernet after 5s: modem init -> WiFi autoConnect -> captive portal
4. WiFi connect/disconnect hooks into conn_mgr so default route and network-availability checks stay accurate throughout the cycle

- For getting MAC address get_device_mac() is used now instead of WiFi.macAddress() in case user doesn't connect to WiFi from the beginning
- getDeviceStatusStamp() reports ethernet / wifi / none with appropriate RSSI (0 for ethernet)
- WifiCaptivePortal.resetSettings() guarded by conn_mgr_is_wifi()


src/api-client/submit_log.cpp
```diff
@@ -27,7 +39,7 @@ bool submitLogToApi(LogApiInput &input, const char *api_url)
 
                     HTTPClient &https = *httpsPointer;
 
-                    https.addHeader("ID", WiFi.macAddress());
+                    https.addHeader("ID", get_device_mac());
```
schrockwell: Is this guaranteed to be the same as `WiFi.macAddress()`? We can't risk this value being incorrect, because the `ID` header is how we identify devices on the server during initial setup.

cerebrumdk: Short answer is yes, they are the same.

There are no detailed info on functions' definitions regarding mac addresses, but basic idea is we have esp_efuse_mac_get_default(), which returns "base MAC address which is factory-programmed by Espressif in EFUSE.", and then we have WiFi.macAddress()  that calls esp_netif_get_mac(), which gets same EFUSE MAC, and generates a mac address of networking interface by adding 0, 1, 2 and 3 to it (depending on the amount of interfaces used). The thing is, the only interface used in terminal was WiFi STA, hence 0 was added to factory-programmed mac addr. But adding 0 doesn't change the number, so in other words WiFi.macAddress() was returning the same thing esp_efuse_mac_get_default() - base mac address of the chip. 

And my get_device_mac() uses esp_efuse_mac_get_default(). 
It's better to use get_device_mac() now, because otherwise we would get 2 different addresses for Ethernet interface and WiFi STA interface

I also asked @GoReNeY to flash his terminal with my PR, and images loaded from server without problems.

https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/network/esp_netif_programming.html#_CPPv417esp_netif_get_macP11esp_netif_tA_7uint8_t

https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/misc_system_api.html#_CPPv425esp_efuse_mac_get_defaultP7uint8_t

https://espressif-docs.readthedocs-hosted.com/projects/esp-idf/en/stable/api-reference/system/system.html#number-of-universally-administered-mac-address

schrockwell: Thank you for confirming - that sounds good to me.

schrockwell:
Hi @cerebrumdk, thanks so much for this contribution. I'm closing this PR because @GoReNeY has separately integrated this work into `main` to support a future TRMNL ethernet solution.

cerebrumdk:
> Hi @cerebrumdk, thanks so much for this contribution. I'm closing this PR because @GoReNeY has separately integrated this work into `main` to support a future TRMNL ethernet solution. 

You are welcome:)

PR #362, opened 2026-05-20
Adding webp file support

ronakjain90:
This PR adds `webp` file support. webp files in some cases leaves a much smaller footprint than PNG (almost 40% smaller file size), this would lead to faster image display and longer battery life.

Tested on TRMNL X


schrockwell:
Per firmware team discussion today, this is cool that it's possible, and we will keep in our back packet if future optimization is necessary.

Due to the high RAM usage of the webp decoder, this would only be supported on X (and future OG gen2) and some BYOD devices. The complexity of this support matrix is not worth it right now, especially in light of all the rendering and caching improvements we have made on the server end in the past few months.

PR #367, opened 2026-05-22
Merge initial support for OG gen2

bitbank2:
This PR adds initial support for the TRMNL OG gen2 PCB (ESP32-C5). Basic TRMNL functionality is working, but additional code is needed to support all of the product features.


schrockwell:
I merged in main, then wired up gen2 to the two new WiFi headers. I extracted a function `getWiFiStatus()` into the new `wifi_network.cpp` as the single source of truth for this info.

PR #370, opened 2026-05-27
Improve USB charging status indication and headers

schrockwell:
- always show charge indicator when USB cable is connected (even it not _technically_ charging)
- add `USB-Connected` header
- add `Image-Cached` header
- add `Wake-Time` header


schrockwell:
Confirmed I got the headers in `/api/display`:

```ruby
{..., "HTTP_IMAGE_CACHED" => "false", "HTTP_WAKE_TIME" => "12479", "HTTP_USB_CONNECTED" => "true", ...}
```

PR #380, opened 2026-06-02
Remove sdkconfig.(env) autogenerated files

schrockwell:
Remove sdkconfig.(env) from repository - they are auto-generated and gitignore'd


schrockwell:
These files are special and have custom values, so we can't delete them.

PR #382, opened 2026-06-02
changed content type used by modem to "application/json"

GoReNeY:


schrockwell (approved):
Confirmed working on X by running the requests through a local HTTP proxy and inspecting the request headers.

Please add a comment to explain what the "AT+HTTPCLIENT=2,1" command means, because it's not obvious for someone reading the code

PR #383, opened 2026-06-03
Extract custom config to sdkconfig.defaults.*

schrockwell:
sdkconfig files are build artifacts and should not be checked into the repository. Per [the docs](https://espressif-docs.readthedocs-hosted.com/projects/esp-idf/en/stable/api-reference/kconfig.html#using-sdkconfig-defaults) (see [also](https://espressif-docs.readthedocs-hosted.com/projects/esp-idf/en/stable/api-guides/build-system.html#custom-sdkconfig-defaults)):

> when sdkconfig file is under revision control, the fact that sdkconfig file gets changed by the build system may be inconvenient. The build system offers a way to avoid this, in the form of sdkconfig.defaults file.

I am having this issue locally, and I suspect others are as well. The `sdkconfig.TRMNL_X` file is constantly being regenerated with a one-line diff, which adds friction every time I perform a Git operation.

Following the recommendation of the docs, I have created `sdkconfig.defaults.*` files which contain our custom config values, and they _are_ checked into the repo. The `SDKCONFIG_DEFAULTS` flag is passed in the build phase so that the configs are derived correctly for each model.

This is a major readability improvement, because it is much more clear which custom config values are applied to each model, compared to before where the custom values were mixed in with all the default values.

---

For the purpose of this PR review, I have regenerated all the model-specific sdkconfig files using the defaults, for inspection, to make sure the diffs are acceptable. Then those files will be deleted and gitignore'd to prevent further churn while still keeping our custom configs.

So, please review these and let me know if any of them are incorrect. Discrepancies can be fixed via the defaults, and then we can delete all of these:

- [ ] sdkconfig.TRMNL_X
- [ ] sdkconfig.TRMNL_X_dev
- [ ] sdkconfig.TRMNL_X_dev_no_qa
- [ ] sdkconfig.TRMNL_X_E1003
- [ ] sdkconfig.TRMNL_X_EPDIY
- [ ] sdkconfig.TRMNL_X_LILYGO_T5PRO
- [ ] sdkconfig.TRMNL_X_PAPERS3
- [ ] sdkconfig.TRMNL_X_SENSORIAC5
- [ ] sdkconfig.TRMNL_X_SENSORIAS3


schrockwell:
Per @GoReNeY 

> I think that change would create more hassle than it solves. Any config change made through menuconfig ends up in sdkconfig.{env}, not in the .defaults file, so we'd have to manually copy those changes into the defaults file every time, which doesn't sound great

PR #388, opened 2026-06-04
Add WiFi-Band and WiFi-SSID headers

### LINKED ISSUE #378: Add RSSI band HTTP header
## Why

With the new support for Wifi bands (i.e. 2GHz, 5Ghz) in Firmware 1.8.5, I'd like to suggest adding a new header to the firmware `/api/display` requests: `Wifi-Band`. This would inform the server which WiFi band the device has been configured to use.

## How

- Add a `WiFi-Band` header.
- Add floats for values. Example: `2.4`, `5`, `6`, etc. These would be strings on the Firmware side but cast into floats on the server side.
- When there is no appropriate WiFi band, then the header to should be omitted.

## Notes

The above would allow each server (Core and BYOS) to display this information within the UI for each device and improve the debug-abilty of the connected devices because it's easy to forgot which band you used to connect the device with to begin with.

schrockwell:
Fixes #378. This PR adds the `WiFi-Band` and `WiFi-SSID` headers to the `/api/display` request.

`WiFi-Band` reports the _currently-connected_ band. This means that a user can configure 5 GHz, but it may fall back to 2.4 GHz on connection failure.

SSIDs may contain non-ASCII characters, which are not compatible with HTTP headers, so we resort to URL-style percent escaping per [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986)

## Examples (captured on the wire)

I have two separate networks set up at home with distinct SSIDs:

```
    WiFi-Band: 2.4
    WiFi-SSID: Hidden%20Orchard%202.4

    WiFi-Band: 5
    WiFi-SSID: Hidden%20Orchard%205
```

## Todo
- [x] test OG
- [x] test X 2.4 GHz
- [x] test X 5 GHz

### DIFF
--- a/lib/trmnl/include/api-client/request_headers.h
+++ b/lib/trmnl/include/api-client/request_headers.h
@@ -19,6 +19,11 @@ HttpHeaderList buildDisplayHeaders(const ApiDisplayInputs &inputs);
 HttpHeaderList buildSetupHeaders(const ApiSetupInputs &inputs);
 HttpHeaderList buildLogHeaders(const ApiLogInputs &inputs);
 
+// Percent-encode a string per RFC 3986 (unreserved chars pass through, all
+// others become %XX). Used to make arbitrary values — e.g. a Wi-Fi SSID with
+// spaces, ':' or control bytes — safe inside an HTTP header value.
+String percentEncode(const String &value);
+
 // Format a header list as a newline-separated "Name: value" string with no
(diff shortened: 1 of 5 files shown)


NiKiZe:
This is a privacy issue if not BYOS, but also the more helpful information should be the BSSID.

ryanckulp:
@NiKiZe i agree, we'll drop SSID.

PR #389, opened 2026-06-04
Add ESP32-M075 GDP075FW1 BYOD target

gibbsoft:
## Summary
- add an `esp32_m075_gdp075fw1` PlatformIO target for ESP32-M075-style boards with the Good Display GDP075FW1 7.5 inch 800x480 four-color panel
- wire the panel into the existing 4-color `EP75YR_800x480` render path and document the display GPIO map
- document the TRMNL BYOD `Waveshare 7.5" B/W/R/Y` profile and working `uvx esptool` flashing commands
- document the NVS erase command needed to force WiFi captive portal setup on this no-button target, including the required hexadecimal `0x` prefixes
- add CI coverage for compiling the new board environment

## Validation
- `pio test -e native`
- `pio run -e esp32_m075_gdp075fw1`
- `pio run`
- `pio check --skip-packages --fail-on-defect high`
- Not rerun after the docs-only WiFi setup note.

## Hardware note
Tested with the WiFi version of the [GDP075FW1/GDP075FU1 7.5 inch 4-color e-paper tag](https://buy-lcd.com/products/gdp075fu1), registered through TRMNL BYOD as `Waveshare 7.5" B/W/R/Y`. That profile delivered four-color content successfully.

The target currently disables button wake and reports a fixed battery voltage until board-specific wake and battery telemetry pins are identified. On the tested board, holding the reset button only resets the ESP32 and does not clear saved WiFi/API settings, so the README documents erasing this target's NVS settings partition (`erase-region 0x9000 0x5000`) to force captive portal setup mode. The `0x` prefixes are required because the erase offset and size are hexadecimal; decimal `5000` is not 4096-byte aligned.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -69,6 +69,9 @@ jobs:
       - name: build (trmnl)
         run: pio run -j ${{ runner.os == 'Windows' && '1' || '4' }} -v
 
+      - name: build (esp32_m075_gdp075fw1)
+        run: pio run -e esp32_m075_gdp075fw1 -j ${{ runner.os == 'Windows' && '1' || '4' }} -v
+
       #- name: build (TRMNL_X)
       #  run: pio run -e TRMNL_X -j ${{ runner.os == 'Windows' && '1' || '4' }} -v
 
--- a/README.md
+++ b/README.md
@@ -251,6 +251,57 @@ There are technical and non-technical options to flashing firmware.
 
 ![Image Alt text](/pics/bin_folder.png "Bin")
 
+### Good Display GDP075FW1 / ESP32-M075
+
+The `esp32_m075_gdp075fw1` PlatformIO environment supports ESP32-M075-style
+controller boards paired with the Good Display GDP075FW1 7.5 inch 800x480
+four-color e-paper panel.
+
+This target has been tested with the WiFi version of the
+[GDP075FW1/GDP075FU1 7.5 inch 4-color e-paper tag](https://buy-lcd.com/products/gdp075fu1).
+
+```bash
+pio run -e esp32_m075_gdp075fw1
+```
+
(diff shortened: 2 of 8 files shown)


bitbank2:
I assume that there are very few people with this board, and of those, few would like to run TRMNL software on it. We encourage people to use the less expansive and more widely available Seeed Studio DIYKit or Xiao ePaper breakout boards. @ryanckulp What do you think?

ryanckulp:
hey @gibbsoft, thanks for this PR. we haven't yet seen demand for this product and the nature of it requiring a USB connection for WiFi connectivity is somewhat against our ethos.

what do you think about maintaining a soft fork, and in exchange we'll link to it from our BYOD docs? if it picks up steam we can reconsider folding it into main.

gibbsoft:
hi @ryanckulp it doesn't need a USB connection for wifi, it does have wifi built in.  USB is just required for initial flashing and if someone wants to reset the wifi connection back to AP fallback mode.

Aside from buying 2 of these and a TRMNL BYOD license just to play with, I don't have any skin in the game. Sadly I can't commit to keeping the fork up to date as I'll no doubt soon get enthused about something unrelated to play with instead. 😄

ryanckulp:
huh ok, i was reading spec docs which said it needs USB-C plugged in for the wifi to turn on. but yeah, i think we want more demand before adding to our maintenance stack. we're at dozens of devices currently and seeking to simplify.

gibbsoft:
Oh... it has an internal battery connector but doesn't ship with a
battery... so it does need USB power to function at all, so that's kind of
true if you look at it like that.

schrockwell:
Thanks again for the contribution @gibbsoft, but we are going to close until we see more demand for this particular screen.

> I'll no doubt soon get enthused about something unrelated to play with instead

Are you… me?

Issue #285, opened 2026-01-24
bug - remembering multiple WiFi networks


ryanckulp:
according to [this review](https://turbulencegains.com/trmnl-review-2026/) we still don't correctly save + cycle through known SSID <> credential pairs upon wakeup.

https://github.com/usetrmnl/trmnl-firmware/blob/fa638ff1fdb5956cf60035aad7f13b451e799e31/lib/wificaptive/src/WifiCaptive.cpp#L228

we've attempted to resolve this twice: 
- https://github.com/usetrmnl/trmnl-firmware/pull/28
- https://github.com/usetrmnl/trmnl-firmware/pull/76

if a non team member wants to take a crack at this sooner than later, we're happy to figure out a bounty (cash/crypto/gear).

donnib:
any progress on this ? is it working now or ?

Walderbash:
Hi All,
Changing to full channel scan will fix this issue, but do so at a cost of having to scan every wake/refresh, making the refresh take longer and wasting battery life on the channel scan. 
My suggestion would be to do a fast connect to the last known access point, and only if that fails, or it has low RSSI, do a full scan and roam to a better AP. This would only be slower the first refresh after you've moved the trmnl to a new location, and would be faster the rest of the time. (assuming most people don't walk around with their TRMNL strapped to their wrist)

I've created pull request https://github.com/usetrmnl/trmnl-firmware/pull/439 that implements this change.
The roaming threshold is set to -75dbm, which should avoid unneccesary roaming even when far away from the AP.

leo1987820:
Hi, I may be able to take a scoped look at this.

Before I start: for the bounty mentioned above, would PayPal be an acceptable cash payout option? Also, since I do not currently have TRMNL hardware locally, would a PR that includes a focused logic fix plus deterministic tests/simulation for the known-SSID selection path be acceptable, with final hardware validation done by a maintainer?

If that path works, I can keep the PR limited to the multi-network credential selection/retry behavior for this issue.

PR #395, opened 2026-06-06
Connect to strongest AP on multi-AP (mesh) networks

ryanckulp:
The arduino-esp32 default scan method is WIFI_FAST_SCAN, which associates with the first AP found for a matching SSID regardless of signal strength. On networks where one SSID is broadcast by multiple APs (mesh/roaming), the device could latch onto a weak/distant AP instead of the nearest one, as reported with logs in issue #285.

Set WIFI_ALL_CHANNEL_SCAN + WIFI_CONNECT_AP_BY_SIGNAL before WiFi.begin() in initiateConnectionAndWaitForOutcome(), which is the single funnel for every connection path (autoConnect last-used, autoConnect fallback cycling, captive portal connect, and portal reconnect). esp-idf now scans all channels and associates with the highest-RSSI AP for the SSID.

Trade-off: a full-channel scan adds ~1-2s per connect vs fast scan.

Refs #285

### DIFF
--- a/lib/wificaptive/src/connect.cpp
+++ b/lib/wificaptive/src/connect.cpp
@@ -180,6 +180,16 @@ WifiConnectionResult initiateConnectionAndWaitForOutcome(const WifiCredentials c
     // always start with a clean state - disable any previous configuration
     disableWpa2Enterprise();
 
+    // Pick the strongest AP when an SSID is broadcast by multiple access points
+    // (mesh/roaming networks). The arduino-esp32 default is WIFI_FAST_SCAN, which
+    // associates with the FIRST AP found for the SSID regardless of signal strength,
+    // so the device can latch onto a weak/distant AP. WIFI_ALL_CHANNEL_SCAN scans
+    // every channel first, then WIFI_CONNECT_AP_BY_SIGNAL connects to the AP with the
+    // highest RSSI. Trade-off: a full-channel scan adds ~1-2s to each connect, which is
+    // an acceptable cost for reliably joining the nearest AP. See issue #285.
+    WiFi.setScanMethod(WIFI_ALL_CHANNEL_SCAN);
+    WiFi.setSortMethod(WIFI_CONNECT_AP_BY_SIGNAL);
+
     wl_status_t beginResult;
 
     if (credentials.isEnterprise)


ucffool:
we've totally had users bit by this and no idea why it was "holding onto an AP far away". This is going to be one of those silent benefits because we will hear less from the growing number of mesh home networks.

schrockwell:
Is there any way we can (quickly) detect a low RSSI before resorting to full channel scan? E.g. if the first AP it finds has an RSSI of > -60 dBm, don't bother with a full scan, since the signal is strong.

I don't know if this would add or subtract to the overall connection time, I guess it depends on the mechanism and timing.

ryanckulp:
> RSSI of > -60 dBm

i like this approach but we should probably be more flexible. around 50% of devices are below this threshold. mine is -67 and i have a mesh network, so i imagine it's connecting to the wrong AP. in which case this PR should help.

```rb
Device.where("rssi < -60").count
=> 15768
Device.where("rssi < -70").count
=> 6040
Device.where("rssi < -80").count
=> 1674
Device.where("rssi < -90").count
=> 140
```

perhaps: review/merge this first, give it a couple days with a new release to re-capture RSSI values across the fleet, then make this optimization?

schrockwell:
Good call - makes way more sense to act on real data, than to blindly guess.

schrockwell:
I found another complication we should consider: 5 GHz networks.

On gen2, the WiFi mode is set to `WIFI_BAND_MODE_AUTO` at boot, which means it will scan both 2.4 and 5 GHz networks, and then pick the strongest out of all of those. It's a longer scan (more channels), and also could result in a different band than what the user selected. We would need to set `WIFI_BAND_MODE_2G_ONLY` or `WIFI_BAND_MODE_5G_ONLY` based on the user's preference.

On X, this change has no effect for 5 GHz networks.

Happy to look into both these issues if you want.

ryanckulp:
good find. i prefer we table that for later because this PR's intent is primarily to resolve "multiple wifi network" support for OGs on the go.

schrockwell:
I will set up a second UniFi AP at home and test this and merge if it looks good.

schrockwell:
Confirmed working on X.

I set up a "Downstairs" access point across the house - same SSID, 2.4 GHz. I configured the TRMNL X Wi-Fi when I was upstairs in my office, where there is also an AP directly next to the device.

Before, it consistently chose the downstairs AP at -80 dBm:

CleanShot 2026-06-12 at 15 08 07@2x

After, it picked the Office at -40 dBm.

CleanShot 2026-06-12 at 15 14 16@2x

PR #398, opened 2026-06-08
Fix E1001 failing to display images

joelhawksley:
Re-initialize the EPD controller (initIO) before each refresh in display_show_image() and display_show_msg(). After a non-blocking refresh with light sleep enabled, the controller can be left busy or in deep-sleep, causing subsequent image writes to silently fail on the Seeed reTerminal E1001. Previously this re-init only ran behind the i426Workaround flag, which is now removed since the re-init always runs.

### DIFF
--- a/src/display.cpp
+++ b/src/display.cpp
@@ -1730,7 +1730,6 @@ void display_show_image(uint8_t *image_buffer, int data_size, bool bWait, bool b
     auto height = display_height();
 //    uint32_t *d32;
     bool bAlloc = false;
-    static int i426Workaround = 0;
 #ifdef BB_EPAPER
     int iRefreshMode = REFRESH_FULL; // assume full (slow) refresh
 #else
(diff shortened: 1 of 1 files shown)


schrockwell:
@bitbank2 Please take a look for me. It's related to `bb_epaper`, and I don't have this device to test.

dmellok:
Tested #398 on a reTerminal E1001. On `main`/v1.8.8 the panel wouldn't display reliably; after building `seeed_reTerminal_E1001` from this branch and flashing, it paints correctly and is stable — no resets across boot + refresh cycles (verified over serial: single `rst:0x1 (POWERON)`, no reboot loop). Thanks for the fix! 🙏

bitbank2:
I don't believe that this is the right fix for this problem. The choice of GPIO lines used on the e1001 seems to be the issue. Light sleep is changing the state of the GPIOs and messing up subsequent updates. Please try disabling light sleep in the platformio configuration for this target instead of adding all of those init() calls @joelhawksley .

joelhawksley:
@bitbank2 disabling light sleep (`	-D DO_NOT_LIGHT_SLEEP=1`) did not fix the issue. Happy to try whatever on my local device, just let me know.

joelhawksley:
@schrockwell rebased. I can confirm that this patch is still needed for the e1001 to work on the current main sha.

bitbank2:
@joelhawksley I am still not convinced that this is the right fix for this issue. The UC81xx controllers don't need constant re-initialization and the panel is put into light sleep, not deep sleep, so it doesn't need to be reset or re-initialized. I don't have this hardware, but I think it would be helpful to get my hands on it for testing.

joelhawksley:
@bitbank2 I have extras I could ship you, but Amazon might be your best route: https://www.amazon.com/reTerminal-E1001-Monochrome-ESP32-S3-SenseCraft/dp/B0FRFHVHNZ

I'm not convinced it's the right fix, but it does work.

bitbank2:
ok, I'll order one. I'll let you know when I receive it.

schrockwell:
@joelhawksley When you get a chance, please rebase/merge from main. Some conflicts were introduced by a major refactor on our end. I don't have push permission on this branch.

bitbank2:
I finally received my E1001 today @joelhawksley . I'm building 1.8.16 firmware on it and cannot reproduce any problems with display update. Can you please retest with the latest firmware? I will also do some power testing and see how much battery life we can expect from it.

joelhawksley:
@bitbank2 I still see the same issue here on my device with the 1.8.16 git tag.

bitbank2:
> @bitbank2 I still see the same issue here on my device with the 1.8.16 git tag.

Did you test by cloning the latest code from main or by using the 1.8.16 release tag? If the later, please try the latest main branch which will shortly be released as 1.8.17. Thanks

joelhawksley:
Yes, it works for me on `main`! Closing ❤️

PR #399, opened 2026-06-09
OTA regression

ronakjain90:
**Issue** 
Found an edge case where when OTA update fail or is skipped and if the image is not present in cache, TRMNL stays at the last displayed screen for the next 24 hrs (when the OTA update is attempted again).

**Reproduction steps**
1. checkout main branch and apply the below patch. The patch mimics the behaviour where OTA was last attempted 2 hrs back and the image if found in cache is deleted.

```
Subject: [PATCH] OTA regression
---
Index: src/bl.cpp
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/bl.cpp b/src/bl.cpp
--- a/src/bl.cpp	(revision 5f0684253bf7bb230987d2dd679cd35ca1993854)
+++ b/src/bl.cpp	(date 1781012199169)
@@ -1428,6 +1428,8 @@
   if (update_firmware)
   {
     uint32_t now = getTime();
+    if (now != 0)
+      preferences.putUInt(PREFERENCES_LAST_OTA, now - 3600); // TODO: remove — repro: OTA skipped (<24h)
     if (now - preferences.getUInt(PREFERENCES_LAST_OTA) >= 24*60*60) {
       Log.info("%s [%d]: Last OTA attempt was > 24h ago, proceeding with download...\r\n", __FILE__, __LINE__);
       if (!checkAndPerformFirmwareUpdate()) {
@@ -1682,6 +1684,14 @@
 
   https_request_err_e result = handleApiDisplayResponse(apiDisplayResult.response);
 
+  // TODO: remove — repro stuck EPD on master: force new-image path (no SPIFFS cache bypass)
+  if (update_firmware && apiDisplayResult.response.filename.length() > 0) {
+    char szTemp[36];
+    fixFileName(apiDisplayResult.response.filename.c_str(), szTemp);
+    filesystem_file_delete(szTemp);
+    status = true;
+  }
+
   if (!status && result == HTTPS_SUCCESS) { // this means we already have this image stored in SPIFFS
       char szTemp[36];
 #if defined( BOARD_X_CLASS ) && !defined(BOARD_SEEED_RETERMINAL_E1003)
Index: include/config.h
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/include/config.h b/include/config.h
--- a/include/config.h	(revision 5f0684253bf7bb230987d2dd679cd35ca1993854)
+++ b/include/config.h	(date 1781011748691)
@@ -3,7 +3,7 @@
 
 #define FW_MAJOR_VERSION 1
 #define FW_MINOR_VERSION 8
-#define FW_PATCH_VERSION 6
+#define FW_PATCH_VERSION 5
 
 // Helper macros for stringification
 #define STRINGIFY(x) #x
```
2. Make sure your device is eligible for firmware `1.8.6`

**Fix:**
set `update_firmware = false;` when OTA update fails and `checkAndPerformFirmwareUpdate` should return `true` only if OTA update was successful.

**Tests**
- [x] TRMNL OG
- [x] TRMNL BWRY
- [X] TRMNL X

### DIFF
--- a/lib/trmnl/include/ota_schedule.h
+++ b/lib/trmnl/include/ota_schedule.h
@@ -1,0 +1,11 @@
+#pragma once
+
+#include <stdint.h>
+#include <persistence_interface.h>
+
+constexpr uint32_t OTA_RETRY_INTERVAL_SECONDS = 24UL * 60 * 60;
+#define OTA_LAST_ATTEMPT_KEY "last_ota"
+
+bool otaAttemptDue(uint32_t now, uint32_t lastOta);
+uint32_t otaLastAttempt(Persistence &persistence);
+void otaRecordAttempt(Persistence &persistence, uint32_t now);
--- a/lib/trmnl/src/ota_schedule.cpp
+++ b/lib/trmnl/src/ota_schedule.cpp
@@ -1,0 +1,19 @@
+#include <ota_schedule.h>
+
+bool otaAttemptDue(uint32_t now, uint32_t lastOta)
+{
+  if (lastOta == 0) return true;
+  if (now == 0) return false;
+  return (now - lastOta) >= OTA_RETRY_INTERVAL_SECONDS;
+}
+
+uint32_t otaLastAttempt(Persistence &persistence)
+{
+  return persistence.readUint(OTA_LAST_ATTEMPT_KEY, 0);
+}
+
+void otaRecordAttempt(Persistence &persistence, uint32_t now)
+{
+  if (now != 0)
+    persistence.writeUint(OTA_LAST_ATTEMPT_KEY, now);
+}
--- a/src/bl.cpp
+++ b/src/bl.cpp
@@ -28,6 +28,7 @@
 #include "api-client/submit_log.h"
 #include <api-client/setup.h>
 #include <special_function.h>
+#include <ota_schedule.h>
 #include <api_response_parsing.h>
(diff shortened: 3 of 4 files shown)


bitbank2:
Thanks Ronak, looks like a good change. The addition of tests is always welcome.

schrockwell:
I rebased off `main` and force-pushed this branch.

This PR does not work on OG. It causes the device to totally lock up. Steps to reproduce:

1. make sure device has OTA updates enabled
2. change `config.h` to an old version e.g. `1.8.5`
3. flash device with this branch with that old version number
4. power cycle device
5. note that the screen is now stuck in the loading state and the device becomes totally unresponsive

IMG_1338

schrockwell:
Replying to myself… the problem I reported actually exists on `main` as well, so it's not a symptom of this branch.

schrockwell:
Confirmed working on OG, BWRY, X.

One problem with this approach is that it puts a 24h throttle on successful OTA updates, too. We will soon be rolling out bleeding-edge dev firmware updates from `main` which could occur more frequently. Maybe this is an okay tradeoff for now. But we should add a concession for that later.

PR #401, opened 2026-06-10
Turned off autoreconnect on both 5Ghz modem and main chip

GoReNeY:
Fixes 1+ minute loading when AP is not available


schrockwell:
Looks simple enough. Will test it by creating and removing temporary SSIDs.

schrockwell:
I can't test this easily, because it involves cutting off an AP in the middle of a connection. But it LGTM.

PR #402, opened 2026-06-10
Add buzzer with two-beep WiFi-reset confirmation

joelhawksley:
Adds a passive-buzzer driver and emits a two-beep pattern when the button is held past BUTTON_HOLD_TIME (WiFi reset threshold), giving audible confirmation without requiring the user to release the button. Buzzer pin/freq are defined for the reTerminal E1001; the API is a no-op on boards without PIN_BUZZER.

I've found this to be a nice feature to have as it otherwise takes a few seconds for the display to refresh into WiFi reset mode.

Note: I've only tested this on the ReTerminal E1001.


schrockwell:
Hi @joelhawksley, please resolve the merge conflicts and I can get this in. I don't have permission to push to your branch. I think the problem is that the username `timeframe` on the repo doesn't match your account `joelhawksley`?

joelhawksley:
@schrockwell done!

schrockwell:
Merged, thank you!

PR #403, opened 2026-06-10
Modem speedup

bitbank2:
This change is specific to the TRMNL X. It skips modem (ESP32-C5) initialization if the user's WiFi credentials are for a 2.4GHz SSID. This saves time and energy during each wakeup cycle.

### DIFF
--- a/src/bl.cpp
+++ b/src/bl.cpp
@@ -75,6 +75,7 @@ const char *szHTTPErrors[] = {
 };
 
 bool pref_clear = false;
+bool bModemNeeded = false;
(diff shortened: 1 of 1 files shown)


schrockwell:
CleanShot 2026-06-12 at 13 49 24@2x

I ran 5 tests on 2.4 GHz and 5 GHz networks, before and after this PR. I found ~2s speedup on 2.4 GHz network refreshes.

bitbank2:
Thanks for testing @schrockwell :)

Issue #369, opened 2026-05-26
Rapid tapping of TRMNL X left/right navigation gesture causes that navigation to stop working


schrockwell:
Seen on firmware 1.8.2 and 1.8.4 on TRMNL X. See the video. Steps to reproduce:

1. have some playlist items cached in memory
2. have the firmware in tap gesture mode
3. rapidly tap on the right side of the touch bar to advance to the next screen
4. keep tapping until the next screen finishes loading
5. notice that right taps are no longer detected
6. left and center function as expected
7. right taps are once again detected after a screen refresh

https://github.com/user-attachments/assets/8ee6e860-2881-4afd-80c7-25379b60aa0f

schrockwell:
After discussing with @ryanckulp it seems this is known behavior. We recalibrate the touch bar every time the device goes to sleep, and if a finger is touching the bar during calibration, it's no good.

Maybe we should skip recalibrate-on-sleep if the device was woken up due to a touch action? That way the chance of a user touching it during calibration becomes very small.

itopaloglu83:
End users are the ultimate testers, no amount of planning can survive the first contact. 

I think this is also because the code base is getting too complicated, maybe it should be split into board variations allowing better view of the process flows for each board. It was quite hard to follow where the code sections were going.

schrockwell:
Fixed by #404

PR #404, opened 2026-06-11
Removing touchbar recalibration before sleep

GoReNeY:


schrockwell:
I tested before/after on my X and confirmed that this fixes #369! 🎉

PR #405, opened 2026-06-11
Fix crash on E1003

joelhawksley:
_Full disclosure: I worked with Claude Opus 4.8 extensively here, due to my lack of comfort in C++. This fix works on my E1003. I am happy to test any proposed changes to this PR to verify._

## Problem

`initiateConnectionAndWaitForOutcome()` in `lib/wificaptive/src/connect.cpp` registers WiFi event handlers with `WiFi.onEvent(...)` and later removes them in three cleanup loops via `WiFi.removeEvent(i)`, where `i` is the **event type** (`ARDUINO_EVENT_WIFI_READY .. ARDUINO_EVENT_MAX`).

However, `WiFi.onEvent()` returns a **registration id**, and `WiFi.removeEvent(id)` matches on that registration id — **not** on the event type. So the removal loops never actually remove the handlers, and they leak.

Each handler captures `&eventData`, a **local on the stack** of `initiateConnectionAndWaitForOutcome()`. Once that function returns, the leaked handlers fire on the next WiFi event — e.g. the `STA_DISCONNECTED` / `STA_STOP` events raised by a later `WiFi.disconnect(true)` — and write `eventData.eventCount++` / `eventData.disconnected` into the reclaimed stack frame, corrupting whatever now occupies it.

On Arduino-ESP32 3.x this reproduced as heap corruption / crashes (`CORRUPT HEAP` / `Cache disabled but cached memory region accessed`) shortly after an image download, when the code path that fetches a new image calls `WiFi.disconnect(true)`.

## Fix

Capture the ids returned by `WiFi.onEvent()` and remove the handlers by those ids at all three cleanup sites (via a small `removeEventHandlers()` helper). No behavior change beyond the handlers now actually being removed.

## Testing

Verified on hardware (Arduino-ESP32 3.x): before the fix, fetching a new image reliably crashed in `~ApiDisplayInputs()` while freeing a `String` whose buffer pointer had been smashed by a stale event handler; after the fix the device cycles through repeated fresh image downloads with no corruption.


schrockwell:
Hi @joelhawksley, thanks for looking into this. I don't have the hardware to test it, but it looks solid. Definitely an improvement.

Just a couple small requests:

1. Please trim down the big comment - we don't need all the contextual history with the explanation of the previous bug, just the purpose of the variable
2. Please fix the merge conflict - a simple "accept both" should do it

I have the fixes locally but I can't get the push to your fork working - I think some permission issue on my end.

joelhawksley:
@schrockwell sounds good! If you have the fixes on your end, you are welcome to land them on your own and I can close this PR. It will be a day or two before I can get to a computer to get it cleaned up otherwise.

lib/wificaptive/src/connect.cpp
```diff
@@ -176,16 +177,34 @@ WifiConnectionResult initiateConnectionAndWaitForOutcome(const WifiCredentials c
...
     WifiEventData eventData;
 
+    // Register WiFi event handlers and remember each registration id.
+    // WiFi.onEvent() returns a registration id, and WiFi.removeEvent() matches on
+    // that id -- not on the event type. Passing the event-type index to
+    // removeEvent() (as was done previously) does not remove these handlers, so
+    // they leak. Because each handler captures &eventData (a stack local), a
+    // leaked handler corrupts the stack when a later WiFi event fires after this
+    // function has returned (e.g. the STA_DISCONNECTED/STA_STOP events raised by
+    // a subsequent WiFi.disconnect()).
```
joelhawksley:
```suggestion
    // Register WiFi event handlers and remember each registration id.
```

joelhawksley:
@schrockwell I was able to get to this today after all, feel free to suggest any other changes if you'd like and I'll accept them.

schrockwell:
Fantastic! Thanks so much for your contribution.

joelhawksley:
@trmnl-bot done!

schrockwell:
Thanks again for fixing this. We found this was also causing crashes in TRMNL X, so your contribution has fixed the problem for many more!

PR #407, opened 2026-06-13
Added 5GHz support for the gen2

bitbank2:
This change enables the TRMNL OG Gen2 (ESP32-C5) to connect to either 2.4 or 5GHz networks, depending on which the user chose in the wifi captive portal.


schrockwell:
Superseded by #519

PR #408, opened 2026-06-15
Resume TLS sessions across deep sleep for trmnl.app

ikraamg:
resumes the tls session across deep sleep for the same-host trmnl.app calls (/api/display and /api/log), so each wake reuses the session instead of doing a full handshake.

- adds ResumableWiFiClientSecure, a WiFiClientSecure subclass whose connect() runs a reduced fork of start_ssl_client() with mbedtls_ssl_set_session/get_session, keeping the session in RTC across deep sleep. read/write/stop are inherited unchanged.
- withHttp takes a resumable flag; display.cpp and submit_log.cpp pass it for the trmnl.app calls. s3 stays on plain WiFiClientSecure.
- gated behind -D TLS_RESUME, on in [env:local]; add it to [env:trmnl] to ship on OG.


schrockwell:
Hey @ikraamg. I merged in `main` and tested this on the OG. It does indeed shave 1 second off the total refresh time, from 8.5 to 7.5 seconds. Nice.

However, there is a problem with it reporting the battery level, and maybe other HTTP headers as well. The dashboard reports the device at 1% battery and it serves up the "Battery Low" screen. I suspect there is some issue with either core or the firmware with how headers are handled with reused TLS sessions?

How long can a TLS session be reused for? Is there some kind of timeout, or is it indefinite? I don't know enough about how this system works. I see there is a fallback if resuming fails, so maybe that is good enough.

Also, worth noting this will not work on 5 GHz on X since that is a totally different path. Not sure if this will be possible to implement in that case - probably not, without recompiling the modem firmware.

ikraamg:
@schrockwell Thanks for reviewing and testing this. 

I have added specs to ensure the headers are flowing through and also flashed this again to test it. The [activity](https://trmnl.com/devices/current/activity_logs) tab is reporting my battery correctly on my side and the headers are also showing up on the redis 24 hour log on the admin device show page.

The TLS session can be reused for 7 days on our server (server default I guess) but is only reused on normal refresh cycles, not on button press as the tmp memory gets erased. And yes, if the TLS fails, it will fallback to the default logic.

Yes, this does not work on 5Ghz. I don't have a TRMNL X yet so I have not experimented with that as yet.

schrockwell:
So the headers look good (I have a local HTTP proxy server to inspect them), but for some reason the `Battery-Voltage` values vary wildly, on both USB and battery power. The values are not complete garbage, they are somewhat reasonable (2.5~4.0V) so it's not a corrupt value. Still investigating...

Example values from consecutive requests (plugged in): 4.20, 4.20, 3.02, 4.83, 3.63, 3.02

On main, they are all consistently 4.8V

schrockwell:
I merged the latest `main` into this branch and now it's working perfectly. I can't make heads or tails of it.

I A/B tested the branches multiple times and always saw the same results. I was testing over insecure HTTP anyway, so it would have NOTHING to do with the TLS path. It must have been some regression introduced earlier that has since been resolved - maybe a dependency.

Anyway, this looks great, thank you. Merging away!

ikraamg:
Awesome! Thanks for moving it through! The differing results between the branches sounds very strange indeed.

schrockwell:
@ikraamg One last thing! I gated the new `ResumableWiFiClientSecure` so it only applies to URLs with `https://trmnl.app` to prevent conflicts with S3, etc. as noted in the comments (but not previously enforced) - sound good?

ikraamg:
Yes, thank you for catching that!

schrockwell:
I am working on getting this up-to-date with the code formatting from main. Then I will fix the build errors.

schrockwell:
Final testing on OG yields a reduction of 0.6 seconds (5.8 to 5.2) per refresh.

PR #409, opened 2026-06-15
Add FW-Commit header

schrockwell:
## Problem
In order to support bleeding-edge dev builds, we can't rely on only the `FW-Version` header since it will be unchanged across many commits to `main`.

Previously, there was a concession for generating a firmware version string with a git hash suffix, like `1.2.3+abc123`. But I could not get it to work at all. I think the Python build hook was not working correctly with esp-idf.

## Solution

I opted for a different approach: a new `FW_COMMIT` definition and `FW-Commit` header. The `FW-Version` will never have a `+` suffix.

Reasoning:

1. I am concerned that changing the `FW-Version` header format could break BYOS that are not expecting that format (for example, `Gem::Version.new("1.2.3+foo")` throws an exception).
2. ~~In Semantic Versioning, `+foobar` is meant to denote _pre-release_ versions. But under our current scheme, the suffix would indicate versions that came out _after_ the X.Y.Z release.~~
3. It's generally useful to have the commit version information for debugging, even for production builds.
4. It will enable team devices on the forthcoming `development` channel to detect if an update is needed.
5. The golden dev firmware binary can be promoted to production without needing to recompile.


bkuhlmann:
:bulb: Use of plus (`+`) denotes metadata, *not* pre-release. Pre-releases are always denoted by a hyphen (`-`) per [specification](https://alchemists.io/articles/strict_semantic_versioning). I've mentioned this before but, in Ruby, `Gem::Version` is absolutely the wrong object to use for this because it's a specialized object for Ruby gems only. Using `Versionaire::Version "1.2.3"`, via [Versionaire](https://alchemists.io/projects/versionaire), gives you a proper version type that adheres to strict semantic versioning.

schrockwell:
@bkuhlmann Thanks for the correction. I literally just read the spec and still messed it up.

I know that `Gem::Version` is not kosher. I was just using it as an example for a BYOS implementation that might not expect the `+` character, and could break. It could be any language or framework.

PR #412, opened 2026-06-16
Fixes problems with Xteink X4 display updates and adds a new target environment for button-only wakeup

bitbank2:
The Xteink X4 was previously having problems updating in 2-bit grayscale mode because of the EPD controller's need to reset after each update. This PR fixes that problem and adds a new target environment (xteink_x4_pwr_btn) which can extend the battery life by disconnecting the battery after each update. The user must press the power button to manually switch to the next plugin in their playlist.


schrockwell:
I merged in `main` and resolved `display.cpp` conflicts to the best of my ability. I don't have the hardware here to test, but the merge looks clean.

PR #413, opened 2026-06-17
Move OTA update code to `FirmwareUpdateService`

ronakjain90:
**Problem**

While fixing #399, I noticed that the firmware update logic is scattered throughtout in bl.cpp
- declare a global bool variable `update_firmware`.
- validate if the `firmware_url` is valid (line: 2177-2185), if not mutate the global variable `update_firmware`
- If Firmware url is valid, create `binUrl` (line 2255-2260)
- If OTA was last attempted more than 24 hrs ago, update firmware (line: 1429-1444)
- If OTA was last attempted less than 24 hrs ago skip firmware update (line: 1429-1444)
- actual firmware update code (line: 3086-3215)
- in the end based on the mutated variable `update_firmware` determine if ESP needs to restart or sleep (line 1522-1526)

mutating global variable `update_firmware` makes the code not readable and error prone.

**Solution**
Create a service `FirmwareUpdateService`, that accepts `update_firmware: bool, firmware_url: string`. This service should run all the validations and determine if the device should do a FW update, and proceed to do a FW update if eligible. Do not introduce any logic changes.

The outcome is that this leads to much better readable code.

**Tests**
- [X] TRMNL X 
- [x] TRMNL OG 
- [x] TRMNL BWRY


src/bl.cpp
```diff
@@ -3078,142 +3054,6 @@ static void resetDeviceCredentials(void)
...
-
-    esp_ota_handle_t ota_handle = 0;
-    esp_err_t err = esp_ota_begin(update_partition, OTA_SIZE_UNKNOWN, &ota_handle);
-    if (err != ESP_OK) {
-      Log.fatal("%s [%d]: esp_ota_begin failed: %s\r\n", __FILE__, __LINE__, esp_err_to_name(err));
-      showMessageWithLogo(FW_UPDATE_FAILED);
-      return false;
-    }
-
-    showMessageWithLogo(FW_UPDATE);
```
ronakjain90, line 3192: This code couldn't be ported over, as `FirmwareUpdateService` now just returns the `failureMessage`. So the end result is that FW update doesn't ends-up hijacking the screen real-estate, but passively tries to download the FW binary in the background and displays the end success / failure message to the end user.

schrockwell:
I love this so much, and I will take a deep dive and test this. Let's not merge it until after 1.8.7 - it's too big of a change when we are close to the release.

schrockwell:
I appreciate the refactor, but there was some pushback in team chat about dropping the `FW_UPDATE` and `FW_UPDATE_SUCCESS` messages. I think they are valuable for user feedback. We don't do firmware updates frequently, so it's not a big interruption. And if we did it silently, the device would be non-responsive with no indication why.

The good news - with the refactor, it was super easy and logical to add them back. :)

Tested on OG and X.

PR #414, opened 2026-06-17
Fix OG / BWRY OTA failure

ronakjain90:
**Problem:**
FW version starting 1.7.6 (I might be wrong with version number), my TRMNL OG and now my BWRY fails to do automated OTA update. The only way for me and other users to get latest TRMNL FW is by flashing the FW manually via the flash page, while this works. It'd be nice to get updates via OTA.

**Investigation**
Watching the serial log, I see that the TRMNL OG & BWRY do not even attempt to download the FW. More debugging reveals that the WiFi is explicitly disconnected in `downloadAndShow`, but FW update happens later in the sequence.

```
WiFi.disconnect(true); // no need for WiFi, save power starting here
```

**Solution**
- Reconnect with WiFi when just before downloading the FW.
- display `showMessageWithLogo(FW_UPDATE);` just before connecting to WiFi, this is important for BWRY as it can take 10s for the screen to refresh, during which I see that BWRY acts buggy and do not update FW.

**Reproduction Steps**
- checkout to current `main`
- change the below code from
```
if (otaAttemptDue(now, otaLastAttempt(preferencesPersistence))) {
```
to 
```
if (otaAttemptDue(now, otaLastAttempt(preferencesPersistence)) || true) {
```
- build + upload the FW as 1.8.5 to receive 1.8.6
- on certain effected device you'd see FW update failure.

**Tests**
- [X] TRMNL OG
- [X] TRMNL BWRY
- [X] TRMNL X

### DIFF
--- a/src/bl.cpp
+++ b/src/bl.cpp
@@ -1217,21 +1217,7 @@ void bl_init(void)
   {
     // WiFi saved, connection
     Log.info("%s [%d]: WiFi saved\r\n", __FILE__, __LINE__);
-#ifdef BOARD_TRMNL_X
-    WifiCredentials lastCreds = WifiCaptivePortal.getLastCredentials();
-    int connection_res;
-    if (lastCreds.is5GHz && g_modem)
-    {
-      Log.info("%s [%d]: 5 GHz network saved — connecting via modem\r\n", __FILE__, __LINE__);
-      connection_res = g_modem->connectToNetwork(lastCreds.ssid, lastCreds.pswd) ? 1 : 0;
-    }
-    else
-    {
-      connection_res = WifiCaptivePortal.autoConnect();
-    }
-#else
-    int connection_res = WifiCaptivePortal.autoConnect();
-#endif // BOARD_TRMNL_X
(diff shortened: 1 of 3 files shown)


schrockwell:
Confirmed this fixes my BWRY, which would previously fail on every OTA update.

src/bl.cpp
```diff
@@ -3149,6 +3135,15 @@ static bool checkAndPerformFirmwareUpdate(void)
   }
 #endif
 
+  showMessageWithLogo(FW_UPDATE);
+
+  if (!ensureWifiConnected())
```
ronakjain90, line 3140: This WiFi reconnection code being a power drain, this would be executed max 1x/day.

schrockwell:
Per meeting discussion today, this is not an optimal fix because the WiFi disconnects and then reconnects, which takes a long time and is a battery drain. But as @ronakjain90 pointed out, this only happens 1x per day, so it's not a major imposition.

This logic will be revisited in #413 where we can be smarter about deciding when to fetch/display images versus firmware updates, and also keep WiFi connected for the duration of the fetch cycle.

PR #415, opened 2026-06-17
Improved battery guage logic to speed up per-wakeup status delays

bitbank2:
This change patches the BQ27xxx library to remove unneeded delays. The result is 8-12 seconds of each wakeup period saved!

### DIFF
--- a/lib/BQ27427/src/BQ27427.cpp
+++ b/lib/BQ27427/src/BQ27427.cpp
@@ -787,22 +787,25 @@ bool BQ27427::writeBlockChecksum(uint8_t csum)
 uint8_t BQ27427::readExtendedData(uint8_t classID, uint8_t offset)
 {
 	uint8_t retData = 0;
-	if (!_userConfigControl) enterConfig(false);
-		
+//	if (!_userConfigControl) enterConfig(false);
+// Adds unnecessary delays since we're just reading data
+
 	if (!blockDataControl()) // // enable block data memory control
 		return false; // Return false if enable fails
 	if (!blockDataClass(classID)) // Write class ID using DataBlockClass()
 		return false;
 	
 	blockDataOffset(offset / 32); // Write 32-bit block offset (usually 0)
-	
-	computeBlockChecksum(); // Compute checksum going in
-	uint8_t oldCsum = blockDataChecksum();
+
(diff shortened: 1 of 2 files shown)


schrockwell:
I see the CI failure reason: https://github.com/usetrmnl/trmnl-firmware/pull/415/changes#diff-6999bb274802bde8d0aec4cb3c439496efceafb8e4f6389efebbbe69a3525eb1R1001-R1004

This new `if` block is indented as if it were supposed to be nested in the parent `else`, but the `else` block is not enclosed in curly braces.

schrockwell:
@bitbank2 Please confirm the braces and logic changes I introduced in this commit are correct? [71b6774](https://github.com/usetrmnl/trmnl-firmware/pull/415/commits/71b6774dd42bcf64e28f47d360eb0c39398b6afe)

schrockwell:
I will test this next week. In order to exercise the non-happy path, the battery needs to be pulled momentarily.

> The chip should hold the golden file in non-volatile memory, but the POR (power on reset) flag gets set when the power drops, so that will cause it to follow the other logic path
> 
> since it's powered from the battery, you would need to unhook the battery to do it

schrockwell:
Pulled the batteries out of my X and tested subsequent resets. Working as expected.

First boot:
```
I: src/bl.cpp [1079]: Connected to BQ27427!
I: src/bl.cpp [1085]: Two batteries detected
BQ27427 golden file [2-cell]: FW version 2.02 confirmed
BQ27427 golden file [2-cell]: CONFIG UPDATE mode active (FLAGS=0x0078)
BQ27427 golden file [2-cell]: all blocks written OK
I: src/bl.cpp [1101]: BQ27427: ITPOR cleared — device in NORMAL mode
```

Second boot:
```
I: src/bl.cpp [1079]: Connected to BQ27427!
I: src/bl.cpp [1106]: BQ27427: ITPOR cleared — device in NORMAL mode
```

schrockwell:
Per @ucffool's recommendation, I tested this on my X on both 2.4 and 5 GHz networks.

- 2.4 GHz: 12s -> 10s refresh
- 5 GHz: 15s -> 10s refresh

So, an improvement in both cases. Nice! But I don't see 8-12 seconds of improvement, that would be pretty impossible considering the total refresh time.

PR #417, opened 2026-06-18
touchbar indicator function

GoReNeY:
Added visual indication for touchbar gestures at the bottom of the screen

### DIFF
--- a/include/display.h
+++ b/include/display.h
@@ -98,6 +98,12 @@ void display_show_battery(float f);
  */
 void display_reset(void);
 
+typedef enum {
+    TOUCHBAR_LEFT,
+    TOUCHBAR_MIDDLE,
+    TOUCHBAR_RIGHT,
+} touchbar_side_t;
+
 /**
  * @brief Function to read the display height
  * @return uint16_t - height of display in pixels
@@ -180,6 +186,8 @@ void display_show_msg_qa(uint8_t *image_buffer, const float *voltage, const floa
  * @param enabled true to enable light sleep, false to disable
  * @return none
  */
+void display_draw_touchbar_indicator(touchbar_side_t side, bool filled);
+
 void display_set_light_sleep(uint8_t enabled);
 
 /**
--- a/src/bl.cpp
+++ b/src/bl.cpp
@@ -493,7 +493,7 @@ static void show_cached_image_by_offset(int offset) {
     if (path.isEmpty()) { Log_info("No cached image for gesture"); return; }
     int file_size = 0;
     buffer = display_read_file(path.c_str(), &file_size);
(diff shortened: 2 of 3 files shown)


schrockwell:
Observations:

1. I only see the left tap gesture if the previous image loads from cache. If it's a full fetch, the bar does not appear.
2. The bar is not visible on plugins that are primarily black - can we add a white border around it, so it stands out in that case?
3. When tapping the center, the center bar re-appears on the loading screen - is that corect? See pic.

IMG_1362

GoReNeY:
@schrockwell 1. Could you describe in which cases your device do a full fetch on a side gesture? 
2. Yeah, it should have a white border ideally, but I'm not sure how to properly do that with partial refresh, asked @bitbank2 to check what can be done with it
3. Yeah, it's intended

schrockwell:
1. Sorry I'm not sure how to reproduce it exactly, but if you tap back & forward enough it will eventually attempt to fetch instead of load from cache. If you think it's too minor we can move forward with it as-is.
2. Understood, thanks. Then let's go with what we have now.
3. Got it.

bitbank2:
sorry for the delay - working on cleanup of the code now...

bitbank2:
I just pushed a cleanup of the drawing logic which shows the rectangle with a white and black outline so that it can be seen on any background. I also fixed the eink power logic (no need to leave it on) and the partial update logic (was not handling the case of drawing the white outline). @schrockwell @GoReNeY

schrockwell:
Confirmed working well on my X. Much more visible on top of a dark image.

In very rare cases, the rectangle does not fill in completely black, it stays white. It feels like the next image starts drawing too quickly, before the black fill has a chance to complete. This can be reproduced by beginning the wi-fi reset process (hold left & right), then tapping to cancel.

It's so fast, I had to take a video and grab a frame.

IMG_1482

bitbank2:
> Confirmed working well on my X. Much more visible on top of a dark image.
> 
> In very rare cases, the rectangle does not fill in completely black, it stays white. It feels like the next image starts drawing too quickly, before the black fill has a chance to complete. This can be reproduced by beginning the wi-fi reset process (hold left & right), then tapping to cancel.
> 
> It's so fast, I had to take a video and grab a frame.
> 
> IMG_1482

I believe this is an inverted color problem, not a "not filled" problem. For some reason @GoReNeY added logic to invert the color of the rectangle in some cases. Was this to indicate different gesture recognition?

GoReNeY:
@schrockwell 
it is not a bug, but intended behaviour, TAP gestures are displayed as white rectangle with black borders, and HOLD gestures are displayed as black rectangles with white borders

schrockwell:
Got it! Thanks. LGTM

PR #419, opened 2026-06-18
aesthetics - move sdkconfigs

ryanckulp:
non functional improvement. i don't like seeing random brand names and i don't enjoy the growing list of files in our repo's root.

image

this PR namespaces sdkconfig* files (1 per platformio environment) to an `sdkconfigs` folder. couple notes from my clanker:

```
## Minor caveat
PlatformIO's "should I reconfigure?" check hardcodes <root>/sdkconfig.defaults. 
With defaults now in the folder, editing it won't auto-trigger a reconfigure — 
touch a per-target config or do a clean build after editing defaults. 
(The per-target files are unaffected; their mtime is still tracked.)

## environment inheritance
env:trmnl_test (which extends env:trmnl) now nominally resolves to …sdkconfig.trmnl_test instead of inheriting …trmnl. It's framework = arduino (no ESP-IDF), so the option is inert for it — zero functional effect, and arguably more correct.
```


ryanckulp:
tagging @schrockwell for your taste and judgment

schrockwell:
I dig it. I think the caveat is okay, since we barely ever edit that edit the defaults file.

I added `trmnl_4clr` and `trmnl_gen2` as well.

PR #421, opened 2026-06-19
Fixed 2-bit gray updates on 4.26 inch panels

bitbank2:
This is a minor fix to enable 2-bit grayscale mode to work properly on the 4.26" 800x480 panels. They require an extra 'panel init' after each update.


schrockwell:
I don't have a device to test this, but I trust it!

Issue #320, opened 2026-03-15
Custom hostname


joostlek:
Hey!

It would be really cool if we can have devices have a custom hostname so we are able to detect the devices on the network better. Currently the hostname for the TRMNL itself is esp32c3-aabbcc, while it would be nicer if we can have trmnl-xxxx.

My main use case for this is being able to detect the TRMNL on the network to discover the TRMNL home assistant integration for them.

schrockwell:
Thanks for the suggestion. This has been implemented in 11f7919 in firmware 1.8.8+

PR #426, opened 2026-06-20
Add friendly ID to network hostname

ronakjain90:
**Problem**
Many modern routers give users a way to manage their devices, view network activity, restrict traffic, whitelist domains etc. TRMNL devices broadcast their client name as default `expressif` / `esp32c3-XXXXXX` which is not an ideal way to identify TRMNL in your network. Especially if you have multiple TRMNL devices with different SKUs.

This PR replaces default `expressif`/`esp32c3` with `TRMNL-ABC123`/`TRMNL-OG-ABC123`/`TRMNL-X-ABC123`/`TRMNL-BWRY-ABC123` to broadcast proper WiFi hostname.

**Before**
WhatsApp Image 2026-06-20 at 14 08 00
WhatsApp Image 2026-06-20 at 14 08 01-3
WhatsApp Image 2026-06-20 at 14 08 01-2
**After**
WhatsApp Image 2026-06-20 at 14 08 01
WhatsApp Image 2026-06-20 at 14 30 32
WhatsApp Image 2026-06-20 at 14 30 31

**Tests**
- [x] TRMNL OG
- [x] TRMNL BWRY
- [X] TRMNL X


ryanckulp:
this could be a good default but bigger picture is customizable hostnames. @Fr3dr1ckson has been working on it and we should see progress next week. ref https://github.com/usetrmnl/trmnl-firmware/issues/320

ronakjain90:
Custom hostname would be really helpful especially for corporate deployments. It would be even better if we can sync the hostname from the web-ui keeping the server being the single source of truth. @Fr3dr1ckson feel free to close this or take over this PR as necessary.

schrockwell:
IMO the hostname should still include a suffix for either the MAC address or Friendly ID, like `TRMNL-X-ABC123`, to differentiate multiple devices on the same network.

ronakjain90:
Thanks for the feedback @schrockwell. Now the device does broadcast the hostname as `TRMNL-X-ABC123`

lib/wificaptive/src/wifi-helpers.h
```diff
@@ -2,31 +2,6 @@
...
-
-  for (const WifiStatusNode &entry : wifiStatusMap)
-  {
-    if (wifi_status == entry.value)
-    {
-      return entry.name;
-    }
-  }
-  return nullptr;
-}
```
ronakjain90, line 32: This was duplicate code which was removed from `.h` file as the same code is present in `lib/wificaptive/src/wifi-helpers.cpp` too

schrockwell:
@ronakjain90 Sorry for more feedback, but I think we have to use the MAC address as the suffix instead, like we do for the WiFi captive portal SSID. The device won't have a Friendly ID initially, not until it successfully hits `/api/setup` for the first time

ronakjain90:
Thanks for the feedback @schrockwell. The change requested itself is not a huge code change just need to replace `friendly_id` with `mac_address` however, I'm trying to figure out more.

1. There are so many computing devices (IoT, phone, tablet, computer) and none expose MAC address in their hostname. Is it not standard or viewed as security issue?
2. When TRMNL has not joined any WiFi network, TRMNL broadcasts its SSID as `TRMNL-12345`. it doesn't broadcast as `TRMNL-MAC_ADDRESS`.
3. Friendly ID as you rightly pointed out will be available only after the `/api/setup` endpoint call, but at that point we can assume that the device has successfully joined the network. presence of Friendly ID in hostname can aid when mass-provisioning the devices inside a corporate network.
4. MAC address in-fact is already viewable to me within my router app (refer the attached screenshot with redacted Mac), so `friendly_id` is just an extra information otherwise not exposed.

I see that #320 is merged and it already implements hostname as `trmnl-eeff`  (`mac.substring(12, 14) + mac.substring(15, 17);`) but to me `friendly_id` seems more usable. #320 may not work with 5G router too.

ronakjain90:
This PR is synced with current `main`.

**Changelog:**

- Prefers custom hostname set by the user in Captive WiFi portal
- Uses `TRMNL-OG` / `TRMNL-X` / `TRMNL-BWRY` model prefix, along with `friendly_id`
- Uses default hostname `TRMNL`, if none is available.
- Works with 5G networks too.

Since TRMNL-X uses two chips for 2.4G and 5G networks, using a MAC address produces 2 unique hostnames in the network for the same device which is not ideal.

schrockwell:
I refactored it a bit. Reaching into `config.h` from libraries is a smell. Instead, now it determines the hostname at the application layer in `src` and then injects that value into the libs.

This is hard to test, because my Ubuquiti router is pretty aggressive at caching the device hostnames, even after trying to force it to forget the old hostname. So even after the change, I usually see the old hostname `esp32c3-F5283C`, but once I was able to see `TRMNL-OG-{ID}` so I assume it is working.

(Sidebar: I uncovered a logic issue with the wifi portal and the user trying to set the hostname to `""` - we don't clear the preference in that case. To tackle in a future PR.)

PR #433, opened 2026-06-23
Adjust gen2 flash size to 16MB

schrockwell:
The gen2 device is an `esp32c5_n16r8` which has 16 MB of flash, larger than the OG with only 4 MB.

This adds a new sdkconfig file that overrides that config value for gen2. The `sdkconfigs/sdkconfig.trmnl_gen2` file is still gitignore'd since it is automatically generated from the `sdkconfig.defaults*` files.

Note! The SPIFFs partition table is still only 4 MB! So we are not taking advantage of all the flash space yet. I leave that to @bitbank2 whenever we get to it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,10 @@
 cmake_minimum_required(VERSION 3.16.0)
 # sdkconfig.defaults lives under sdkconfigs/ (per-target configs are redirected there
 # via board_build.esp-idf.sdkconfig_path in platformio.ini). Must be set before project().
-set(SDKCONFIG_DEFAULTS "${CMAKE_SOURCE_DIR}/sdkconfigs/sdkconfig.defaults")
+# Only set the default when an env hasn't supplied its own SDKCONFIG_DEFAULTS chain
+# (via board_build.cmake_extra_args), so board-specific overrides can layer on top.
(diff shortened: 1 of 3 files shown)


bitbank2:
Updated with the X's 16MB FLASH partition layout.

schrockwell:
> Updated with the X's 16MB FLASH partition layout.

Sorry, where's this?

bitbank2:
> > Updated with the X's 16MB FLASH partition layout.
> 
> Sorry, where's this?

In the platformio.ini file - see "board_build.partitions"

schrockwell:
> > > Updated with the X's 16MB FLASH partition layout.
> > 
> > 
> > Sorry, where's this?
> 
> In the platformio.ini file - see "board_build.partitions"

Sorry I should have been clearer, I was looking for a relevant commit. I don't see it on this branch or on main.

bitbank2:
Sorry, I pushed it to the branch used in this merge:
https://github.com/usetrmnl/trmnl-firmware/commit/1d319afff9c8761e8fe90fef2e1943c527d749f0

schrockwell:
Ah I see. GitHub deletes remote branches when PRs are merged, and so you pushed up a new branch and the PR didn't track it, so I didn't see it here.

In the future, please open a new PR off main instead, even for one-liners. It also makes it easier to track which issues solved by which PR.

Issue #235, opened 2025-09-26
Add TLS1.3 Support to Firmware


PatrickTCB:
The trmnl firmware doesn't support TLS 1.3 currently. 

Mbed TLS is the library used for TLS connections and TLS 1.3 needs to be explicitly enabled using MBEDTLS_SSL_PROTO_TLS1_3 in 'sdkconfig.defaults'.

I got this parameter from the Mbed TLS documentation here: https://github.com/Mbed-TLS/mbedtls/blob/development/docs/architecture/tls13-support.md.

schrockwell:
Hi @PatrickTCB, thanks for the docs. This looks like a simple addition - I will try it out and let you know.

schrockwell:
So unfortunately, for OG and similar devices we are currently using the framework espressif32 v6.12.0 which includes a version of Mbed TLS that is too old for TLS 1.3. Updating to a new framework would be a large amount of work that I am not sure we are ready to tackle right now. @GoReNeY what do you think?

I am now looking into supporting this on TRMNL X - see PR #434.

GoReNeY:
We already had a try to switch from espressif32 v6.12.0 but encountered a bunch of issues and did a rollback. I think it's up to @Fr3dr1ckson to decide if we want to try again

PR #434, opened 2026-06-23
Add TLS 1.3 support to TRMNL X

schrockwell:
Addresses #235.

I used `pio run -e TRMNL_X -t menuconfig` to enable TLS 1.3 support, then manually diff'ed those changes over to the other `TRMNL_X_*` environments.

I did not include OG, since it's on espressif32 v6.12.0 which contains a version of Mbed TLS that is too old to support TLS 1.3. We won't make that upgrade at this time.

I also deleted and ignored all the `sdkconfig.*.old` files

## Todo

- [x] add for 2.4 GHz
- [x] test on X 2.4 GHz
- [ ] build custom modem firmware with TLS1.3 support
- [ ] add for 5 GHz (ESP-AT modem)
- [ ] test on X 5 GHz

## Verification

I set up a Caddy reverse proxy with only TLS 1.3 termination, and pointed the TRMNL X at that address.

Here's that Caddyfile for future reference:

```
{
	# Enable global debug logging (handshakes, proxy decisions, errors)
	debug
}

192.168.1.10:8888 {
	# Bind the listener to all interfaces.
	bind 0.0.0.0

	# Issue a local cert via Caddy's internal CA, and only allow TLS 1.3
	tls internal {
		protocols tls1.3
	}

	# Per-site access log: every request + response, written to stdout as
	# human-readable console lines at DEBUG verbosity.
	log {
		output stdout
		format console
		level DEBUG
	}

	# Reverse proxy to the upstream, rewriting the Host header so trmnl.app
	# serves the correct vhost / TLS SNI.
	reverse_proxy https://trmnl.app {
		header_up Host trmnl.app
	}
}
```


schrockwell:
For 5 GHz connections on TRMNL X, this is a bigger lift. We use the ESP32C5 with the [ESP-AT](https://docs.espressif.com/projects/esp-at/en/latest/esp32c5/AT_Binary_Lists/esp_at_binaries.html) firmware as the 5 GHz modem. Currently it is `5.1.0.0-dev`, and I'm not sure where that came from since it's not a production release (latest is 5.0.1.0).

To support TLS 1.3, we'll need to reconfigure and rebuild the modem firmware with  `CONFIG_MBEDTLS_SSL_PROTO_TLS1_3=y` and related settings.

schrockwell:
This is too big of a lift right now. OG requires a major toolchain update which is a big refactor, and X 5 GHz will require custom modem firmware, which we cannot currently flash OTA.

PR #438, opened 2026-06-24
Fixes 2BPP gray shades being darker than expected.

ronakjain90:
**Problem:**
On TRMNL X device the displayed 2 bit image is much darker than in the source file, this makes the gray transition not smooth, producing image that looks distorted.

**Solution**
2-bit PNGs on TRMNL X now decode as 4-bit internally so they use the tuned gray matrix.


schrockwell:
@bitbank2 Could you please take a look? This is particular solution may not be ideal, but it is a proof-of-concept that shows a major improvement.

ronakjain90:
@bitbank2 - @schrockwell asked me to upload the before & after photos of the device after this change. 

### **Reproduction steps**
1. Add any photo based plugin to your playlist.
2. Set the screen to be 2 bit in playlist schedule.
3. Let the device display the image.


### **Before** -- current `main`
WhatsApp Image 2026-06-29 at 12 26 09

### **After** -- current branch
WhatsApp Image 2026-06-29 at 12 26 08

bitbank2:
working on it...

bitbank2:
I tried to adjust the default 2-bit update in FastEPD for the X, but the 10.3 panel has "special needs", so this idea of using the slower, 4-bit custom update pattern designed specifically for that EPD is unfortunately needed. I did a rewrite of the AI slop and added better comments.

ronakjain90:
Thank you @bitbank2 for looking into this.

The earlier test image appears to look fine, but when I tried testing with a zoomed out image the result is pixelated.

Commit (2700b70fbba2d27a60481b0bb5ee9869045e2f6b):
IMG20260705131747 jpg

Commit (7b86af5dfa2ec503e98422a10f6dcf866ecbe8df): 
IMG20260705131347 jpg

ronakjain90:
The pixelation problem appears fixed by the following code

```
static void Expand2bppLineTo4bpp(const uint8_t *src, uint8_t *dest, int width)
{
    uint8_t b;
    uint16_t u16;

    for (int x = 0; x < width; x+=4) {
        u16 = 0;
        b = *src++;
        for (int j=0; j<4; j++) {
            u16 <<= 4;
            u16 |= (((b >> 6) & 3) * 5);
            b <<= 2;
        } // for j
        *dest++ = (uint8_t)(u16 >> 8);
        *dest++ = (uint8_t)(u16 & 0xff);
    } // for x
} /* Expand2bppLineTo4bpp() */
```

PR #439, opened 2026-06-24
WiFi fast connect using cached BSSID + channel

Walderbash:
### Problem

On every wake from deep sleep the device runs a full `WIFI_ALL_CHANNEL_SCAN`, which probes every channel before associating. This adds ~1–2 s to every refresh cycle even when the device is sitting next to the same router it has always used.

### Solution

After each successful connection the device now saves the connected BSSID and channel to NVS. On the next wake it attempts a *fast connect* first: `WiFi.begin(ssid, pass, channel, bssid)` with `WIFI_FAST_SCAN`, which skips the channel scan and associates directly with the known AP. The full channel scan is kept as a fallback.

### Fast connect fallback conditions

The device abandons the fast connect and falls back to `WIFI_ALL_CHANNEL_SCAN` if:

- **AP not found** — the cached BSSID/channel is stale (router replaced, channel changed). Fast connect is capped at 5 s to bound the penalty; `WL_NO_SSID_AVAIL` also causes an immediate exit rather than waiting out the timeout.
- **Weak signal** — RSSI after connecting is below −75 dBm. The device disconnects and re-scans all channels to find a stronger AP (roaming/mesh networks). The new BSSID/channel is saved for next time.

### Behaviour on first boot / existing devices

Devices with saved credentials but no cached BSSID/channel skip the fast connect gate entirely and go straight to the full scan. The hint is saved on the first successful connection, so fast connect activates from the second wake onward. No migration or reset needed.

### Testing

Tested on device. Connection time is consistently **~2 seconds faster** than before. Roaming behaviour (weak signal triggering a rescan that picks up a closer AP) works correctly — the device re-associates and updates the cached hint for next time.

### Worst-case timing

| Scenario | Before | After |
|---|---|---|
| Normal wake (same AP, good signal) | ~2–3 s | ~1 s |
| Weak signal / roaming | ~2–3 s | ~1 s + rescan |
| Stale cache (AP replaced/moved) | N/A | ≤5 s + rescan |
| No cache yet (first wake) | ~2–3 s | ~2–3 s (unchanged) |


schrockwell:
This looks great - thank you! I will review and test it.

Walderbash:
I've added 4 extra fixes, making the full refresh take around 6 seconds:

- we check if the BQ27427 config is still valid, and only apply it if needed.
- we only get the battery designEnergyScale once and afterwards cache it (this can never change without a full powerdown)
- we don't do a full epaper refresh for the loading screen logo. This is a choice, but it does save quite some time and it still looks ok.
- Claude saw a HTTP client crash, moved the wifi client disconnect to the end of the image loading code.

 
10:16:45.407 > I: src/bl.cpp [3279]: total awake time - 5795 ms
10:16:45.408 > I: src/bl.cpp [3280]: time to sleep - 309

lib/wificaptive/src/connect.cpp
```diff
@@ -320,6 +333,48 @@ WifiConnectionResult initiateConnectionAndWaitForOutcome(const WifiCredentials c
...
+                Log_info("WiFi: Fast connect begin, status %s", wifiStatusStr(beginResult));
+                auto fastResult = waitForConnectResult(WIFI_FAST_CONNECT_TIMEOUT);
+                if (fastResult == WL_CONNECTED)
+                {
+                    int32_t rssi = WiFi.RSSI();
+                    Log_info("WiFi: Fast connect succeeded, RSSI %d dBm", rssi);
+                    if (rssi >= WIFI_FAST_CONNECT_MIN_RSSI)
+                    {
+                        for (int i = ARDUINO_EVENT_WIFI_READY; i < ARDUINO_EVENT_MAX; i++)
+                            WiFi.removeEvent(i);
```
schrockwell: Please see #405 for the correct implementation of this event handler cleanup loop.

schrockwell (changes requested):
I'd like to see one more cache-busting condition to force a full scan once every 24 hours.

This will ensure that if the wireless environment changes (new AP installed , or device moves), that it will occasionally find the new strongest signal.

Please also see the note about the WiFi event handlers.

lib/wificaptive/src/connect.cpp
Walderbash: - added the code to remove eventhandlers for the fast-connect wifi calls.
- also added the 24 hour fallback, it will save the epoch time it last did a fullscan and compare that with current epoch to decide to do the fastscan. This does add quite a bit of code, I'm a bit on the fence if this change is worth it; it will solve the device sticking to a good enough AP when a strong AP is available, but any wifi device (phone/laptop) would have that issue, and is something is better solved by making the APs overlap less. (basically I would advise to only do the eventhandler fix)

I've compiled tested this on my X, ran a soak test for 20 hours with fallback scan set to 1 hour and extra logging to the server, saw the fallback scan happen, and could see it roam on low signal strength.

schrockwell:
This is fantastic. I merged in `main` and extracted out the `tryFastConnect()` function which made the logic simpler and more readable.

On X, I measured ~2 seconds of speedup, and over 3 seconds of improvement on OG. That's over 25% reduction of time spent in the most power-hungry device state. This will greatly improve battery life for everyone.

The partial-refresh is neat too. The team loves it. :)

schrockwell (approved):

PR #441, opened 2026-06-25
use `idf_component.yml` to produce consistent production build

ronakjain90:
**Problem:**
platformio automatically grabs the latest package because of carat ^ operator at the start of version number, so it took liberty to get the latest package satisfying the ^1.1.0 not the exact version. This produces inconsistent production build, thus producing classic "it doesn't work on my device" / "my device has this issue" .

without `idf_component.yml`, it produces the below patch
```
Index: dependencies.lock
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/dependencies.lock b/dependencies.lock
--- a/dependencies.lock	(revision 9105d344b84db117ed0804ccf124023c32c24aa8)
+++ b/dependencies.lock	(date 1782387976088)
@@ -12,7 +12,7 @@
   esp-serial-flasher:
     dependencies: []
     source:
-      path: C:\Work\trmnl\firmware\components\esp-serial-flasher
+      path: /Volumes/ExternalHD/Code/TRMNL/firmware/components/esp-serial-flasher
       type: local
     version: 1.11.0
   espressif/cbor:
@@ -66,7 +66,7 @@
       type: service
     version: 1.0.18
   espressif/esp-sr:
-    component_hash: e89fce9c0afe1b6782a9fbc65b5fc2ba5841ed4a39704d08414aad891ccce1a3
+    component_hash: aa58d6a13a49600314a50a27240ccbd17e407f608216a3086d4868e0da2e8053
     dependencies:
     - name: espressif/cjson
       registry_url: https://components.espressif.com
@@ -86,7 +86,7 @@
     source:
       registry_url: https://components.espressif.com/
       type: service
-    version: 2.4.4
+    version: 2.4.6
   espressif/esp-zboss-lib:
     component_hash: 321883d142421f65009972408287441794250057668a11abbdfd8bec77c3309a
     dependencies:
@@ -160,7 +160,7 @@
       type: service
     version: 1.2.2
   espressif/esp_modem:
-    component_hash: f4fa6dab2496af2673a68881132bb7ce0f05bf62e30e5bd0d45b607066378a28
+    component_hash: 475c2185273ddedff2bb0d21a8ab196a9d0ef60595f73d634edf2d8be03a5f54
     dependencies:
     - name: idf
       require: private
@@ -168,7 +168,7 @@
     source:
       registry_url: https://components.espressif.com/
       type: service
-    version: 2.0.1
+    version: 2.0.2
   espressif/esp_rainmaker:
     component_hash: f6fe458fc7a0102ee2879f0247da4b41419e6c07de12031f66e5e9454d25baaa
     dependencies:
@@ -292,7 +292,7 @@
       type: service
     version: 1.0.22
   espressif/mdns:
-    component_hash: 8bcf12e37c58c1d584aef32a02b92548124c7a3a9fcf548d3235c844a035e0f0
+    component_hash: 291732b5df4d181eb75eba9042799a508c4c124503a0c0573132a0a48d2e3fbe
     dependencies:
     - name: idf
       require: private
@@ -300,7 +300,7 @@
     source:
       registry_url: https://components.espressif.com/
       type: service
-    version: 1.11.1
+    version: 1.11.2
   espressif/network_provisioning:
     component_hash: ef2e10182fd1861e68b821491916327c25416ca7ae70e5a6e43313dbc71fe993
     dependencies:
@@ -350,11 +350,8 @@
     - linux
     version: 2.4.0
   espressif/usb_host_ch34x_vcp:
-    component_hash: cc6f17936a11aa0ba460e716bd9b70fcd445824bcc7882afbe3ede3513f80c59
+    component_hash: bf5e9c1c4e6e0a617cc268978b7ae4537d559e6c4b3bdae857dd05b703b4283c
     dependencies:
-    - name: idf
-      require: private
-      version: '>=4.4'
     - name: espressif/usb_host_cdc_acm
       registry_url: https://components.espressif.com
       require: public
@@ -362,7 +359,7 @@
     source:
       registry_url: https://components.espressif.com/
       type: service
-    version: 2.2.0
+    version: 2.2.1
   espressif/usb_host_cp210x_vcp:
     component_hash: c0404c598e2620fe88c35b5bf24d8e3b3ded41d9946fc865843be7f2f75da40e
     dependencies:
@@ -382,7 +379,7 @@
       type: idf
     version: 5.5.2
   joltwallet/littlefs:
-    component_hash: 29eff7921c4e8bce021aeb7fff2723bfcafed661756de700646ea0f6223c2a32
+    component_hash: 01faa388c6305fe77fd6962cda6171e753033458f24f8fae9e9d932e3d817a4c
     dependencies:
     - name: idf
       require: private
@@ -390,7 +387,7 @@
     source:
       registry_url: https://components.espressif.com/
       type: service
-    version: 1.21.1
+    version: 1.22.1
 direct_dependencies:
 - chmorgan/esp-libhelix-mp3
 - espressif/cbor
@@ -414,6 +411,6 @@
 - espressif/usb_host_cp210x_vcp
 - idf
 - joltwallet/littlefs
-manifest_hash: 45774476908c437a0cc6e2e11b25bbac17566aa759e578b6b41dc5131975bbdb
+manifest_hash: 3057f3b1b68a48b45c4055ae9ebed2a73c16b4854890a864b2a543b2237f9e7c
 target: esp32s3
 version: 2.0.0
```

**Solution:**
- use `idf_component.yml` to produce consistent production build
- Introduce cli commands to improve developer productivity.
- delete `dependencies.lock` from git, add the same to `gitignore` because it produces different lock file between `esp32c3` and `esp32s3`
- use CI to compare the `dependencies.lock` file against saved file and warn against potential version drift.


schrockwell:
Is there a way to just check in a lockfile? Isn't that the whole point of lockfiles, to prevent this version drift issue? 

I'm not blaming you of course, but something is clearly failing with the way we have the tooling set up, and ideally I'd like to take a simpler path if it's possible.

ronakjain90:
TRMNL OG & BWRY produces a different `dependencies.lock` file when compared to TRMNL X. I think since we intentionally do not check-in OG/BWRY dependencies.lock we don't see the issue. In short the current `main` doesn't highlight the discrepancy of building multiple devices. 

I also explored a way to create multiple named `dependencies.lock` prefixed appropriately, but that's not currently possible.

dependencies.lock.esp32s3
```diff
@@ -394,6 +394,7 @@ dependencies:
 direct_dependencies:
 - chmorgan/esp-libhelix-mp3
 - espressif/cbor
+- espressif/dl_fft
```
ronakjain90, line 397: this is added as `direct_dependencies` because otherwise it's resolved to `0.5.0` instead of `0.4.0`

dependencies.lock.esp32s3
```diff
@@ -12,7 +12,7 @@ dependencies:
   esp-serial-flasher:
     dependencies: []
     source:
-      path: C:\Work\trmnl\firmware\components\esp-serial-flasher
+      path: components/esp-serial-flasher
```
ronakjain90, line 15: relative path is a newer feature of [IDF Component Manager](https://github.com/espressif/idf-component-manager/releases) manager `3.x`, it's not available in `2.x`. This will make sure that there is no conflict when building from MacOS vs Windows vs Linux.

dependencies.lock.esp32c3
```diff
@@ -0,0 +1,9 @@
+dependencies:
```
ronakjain90, line 1: This is `dependencies.lock` for TRMNL OG/BWRY

dependencies.lock.esp32s3
```diff
@@ -414,6 +415,6 @@ direct_dependencies:
 - espressif/usb_host_cp210x_vcp
 - idf
 - joltwallet/littlefs
-manifest_hash: 0f7cfff62ca57d0d29f698457da843026e779e4ff6b25c17513af0ad6a745950
+manifest_hash: 4547f754dacaa852dfde4d11869ba301405f180f2902c988379fe33c26f66f1d
 target: esp32s3
-version: 2.0.0
+version: 3.0.0
```
ronakjain90, line 420: I updated the components manager using
```
~/.platformio/penv/bin/uv pip install --python ~/.platformio/penv/.espidf-5.5.2/bin/python "idf-component-manager==3.0.3"
```

ronakjain90 (commented):
@schrockwell, you were right for the pushback. I re-read the [documentation](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/tools/idf-component-manager.html) and I found the below

> You may set the build property DEPENDENCIES_LOCK to specify the lock-file path in the top-level CMakeLists.txt. For example, adding idf_build_set_property(DEPENDENCIES_LOCK dependencies.lock.${IDF_TARGET}) before project(PROJECT_NAME) could help generate different lock files for different targets.

So now I'm using `IDF_TARGET` to generate target specific `dependencies.lock`

dependencies.lock.esp32s3
schrockwell: I added `scripts/pin_component_manager.py` as a pre-script to X and related environments. It will automatically install idf-component-manager 3.0.3 so that the user doesn't need to run this manually, and builds should continue to Just Work out-of-the-box without generating any diffs.

schrockwell (approved):

PR #410, opened 2026-06-16
Add Seeed reTerminal E1004 support

limengdu:
## Summary

Adds firmware support for the Seeed reTerminal E1004, a 13.3 inch 1200 x 1600 dual-chip Spectra 6 ePaper device.

Included changes:

- Add the `seeed_reTerminal_E1004` PlatformIO environment.
- Add the E1004 pin map, wake button, battery measurement switch, and UART logging path.
- Route E1004 rendering through the `bb_epaper` Spectra 6 driver path.
- Decode PNG content with the existing Spectra 6 nearest-color palette mapping instead of forcing black/white output.
- Disable `bb_epaper` light sleep before blocking refresh calls on this target to avoid RTC watchdog resets during long E1004 refreshes.
- Fix the Seeed post-build merge helper to invoke PlatformIO's packaged `esptool.py` from the active environment.

## Device parameters for hosted TRMNL support

The hosted TRMNL app will also need a Device Model / palette entry for this hardware. The firmware sends this model header:

- `Model`: `reTerminal E1004`

Suggested device parameters:

- Product: Seeed reTerminal E1004
- Display panel: 13.3 inch E Ink Spectra 6
- Panel model: T133A01
- Resolution: 1200 x 1600 pixels
- Colors: 6
- Palette: Spectra 6
- Image format: PNG
- Firmware panel enum: `EP133A_SPECTRA_1200x1600`
- Panel topology: dual-chip / dual-controller panel
- MCU board target: Seeed XIAO ESP32S3 / ESP32-S3
- Wake button: GPIO4, active low
- UART log bridge: RX GPIO44, TX GPIO43, 115200 baud
- Battery measurement load switch: GPIO21, active HIGH

Without a hosted Device Model / palette entry for this combination, BYOD registration currently fails in the TRMNL web app with `Palette is invalid`.

## Dependency

This firmware target depends on `bb_epaper` support for `EP133A_SPECTRA_1200x1600`. Until that is available in an upstream library release, this PlatformIO environment points at the E1004 support branch:

- `https://github.com/limengdu/bb_epaper.git#codex/e1004-t133a01`

Related `bb_epaper` PR:

- https://github.com/bitbank2/bb_epaper/pull/32

## Root cause fixed

[...]

Issue #342, opened 2026-04-27
Support for SHT40 temperature/humidity sensor


petradios:
I was wondering if there any plans for native support of the SHT40 temperature/humidity sensor that is included on seeed's reTerminal E1001?
After I forked the repository and used an Adafruit dependency I managed to get temperature readings and display it on a serial monitor but still could not manage to get it through an API request (tried with both Terminus and Larapaper). I wonder if I'm doing something wrong on my part or if the way I implemented the http inject was somehow different than the way it is already implemented for the supported sensors. I can provide the code if necessary and thank you in advance.

PR #445, opened 2026-06-28
Add Seeed reTerminal E1004 support (rebased onto main + fixes, based on #410)

oetiker:
Builds on @limengdu's #410 (Add Seeed reTerminal E1004 support), which currently
conflicts with `main`. I rebased it onto current `main` (resolving the conflict)
and folded in a few fixes/cleanups. The E1004 is the Seeed reTerminal 13.3"
(T133A01, dual-chip Spectra6).

### On top of #410
- **Reproducible deps**: pull the EP133A-capable `bb_epaper` from a *commit-pinned*
  fork instead of a mutable branch.
- **Dependency refactor**: introduce `[deps_app_base]` (device deps minus
  `bb_epaper`) so a board can pick its own `bb_epaper` variant without duplicating
  the list — and keep `bb_epaper` out of the native test build.
- **Light-sleep helper**: extract `display_apply_light_sleep_setting()` to replace
  the repeated inline `DO_NOT_LIGHT_SLEEP` blocks, and apply it on the message/QA
  refresh paths that were missing it.
- **post-build**: invoke `esptool.py` directly via `$PYTHONEXE` + the resolved
  `tool-esptoolpy` dir instead of `pio pkg exec` (more robust in CI; affects all
  Seeed XIAO E100x merges).
- Dual-controller display init (`setCS2`), panel-power enable, wake-button RTC
  config, battery read, and dev serial pins.

### Testing
Built and flashed to a reTerminal E1004 — boots, fetches, and renders on the
13.3" Spectra6 panel.

Credit to @limengdu for the original E1004 bring-up in #410.

---

### Update: onboard SHT4x sensor reporting

Also wires up the E1004's **onboard Sensirion SHT4x** (temperature + humidity)
and reports it to the backend via the existing `SENSORS` header — no server-side
changes needed. Verified on a real reTerminal E1004:

```
SENSORS: make=Sensirion;model=SHT4x;kind=temperature;value=29.77;unit=celsius;created_at=…,
         make=Sensirion;model=SHT4x;kind=humidity;value=55.5;unit=percent;created_at=…
```

- Pure, native-unit-tested CRC-8 + conversion core in `lib/trmnl/` (5 tests).
- Self-contained `Wire` reader (`0xFD`, 6 bytes, CRC-gated) on the SHT4x's I²C
  bus (`GPIO19/20`, addr `0x44`); all paths guarded by `HAS_ONBOARD_SHT4X` so
  other boards are unaffected.
- `created_at` stamped at send time so it's valid after SNTP sync.

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -46,7 +46,7 @@
 
 #define DISPLAY_BMP_IMAGE_SIZE 48062 // in bytes - 62 bytes - header; 48000 bytes - bitmap (480*800 1bpp) / 8
 #define DEFAULT_IMAGE_SIZE 48000
-#if defined(BOARD_X_CLASS)
+#if defined(BOARD_X_CLASS) || defined(BOARD_SEEED_RETERMINAL_E1004)
 #define MAX_IMAGE_SIZE 750000 // Use PSRAM on the ESP32-S3 (all X-class boards have PSRAM)
 #else
 #define MAX_IMAGE_SIZE 90000 // largest compressed image we can receive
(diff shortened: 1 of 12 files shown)


schrockwell:
Thanks for this contribution! I am working on getting the E1004 to test this out. Will keep you updated.

oetiker:
it works very well :)

photo_5916022717159247075_y

schrockwell:
I am still working on getting the hardware to test - thanks for your patience.

goldenresponse:
FWIW just 2 cents from the peanut gallery, I tested this on a newly-arrived e1004 device last night, and at least in my testing it works well. I am using this as essentially a digital picture frame via the Alias plugin w/server on LAN.

schrockwell:
I got the device in today. Will work on merging in the formatting changes from `main` and testing for v1.8.12.

The SHT4x support may also solve #342 - TBD.

ururk:
Got my device today - flashing worked (did not try `_dev`). Some comments:
- I forgot to put in a custom server config - none of the buttons would get it into setup mode and I had to erase the device and reflash
- While the device has three buttons (six if you include the mirrored back buttons, which may or may not map to a different set of io pins), only the left arrow works to refresh the screen. Feels like it should support all three (maybe?) or just the refresh icon one instead of the arrow
- This is a UI comment which may not be specific to this device, but there is no indication the frame is loading a new screen. If this is normal/expected, then ignore :D but even a small solid dot or shape would be neat to have.

schrockwell:
Hi folks. We appreciate all the hard work and testing that has gone into this. The upstream `bb_epaper` library is adding support for this display, which will benefit many more users beyond just TRMNL. We have opted to rely on the dependency, like we currently do for many other displays. Thanks again.

robbiet480:
[FYI the `bb_epaper` PR was rejected](https://github.com/bitbank2/bb_epaper/pull/32)

PR #446, opened 2026-06-28
reTerminal E1001/E1002: wake on the extra front buttons (next screen)

tajchert:
## What

The Seeed reTerminal E1001/E1002 have three front buttons, but the firmware only used the green button (`GPIO3`) as a deep-sleep wake source. The two white buttons — Left (`GPIO4`) and Right (`GPIO5`) — were unused.

This wires both white buttons to act as a **"next screen"** control: pressing either wakes the device and runs a normal refresh, which advances to the next playlist item. The green button keeps its existing behavior (refresh / double-click special function / long-press WiFi reset).

## How

- On these boards the ESP32-S3 deep-sleep wake source is switched from single-pin `ext0` to **`ext1` with `ESP_EXT1_WAKEUP_ANY_LOW`** over `GPIO3 | GPIO4 | GPIO5` (with RTC pull-ups), so any of the three buttons can wake the device.
- On wake, the source pin is read back via `esp_sleep_get_ext1_wakeup_status()`. A white-button wake skips the green-button press classification (and its double-click wait) and falls straight through to the normal display refresh.
- All of this is gated behind a new `HAS_NEXT_BUTTONS` macro defined only for `BOARD_SEEED_RETERMINAL_E1001` / `_E1002`, so every other board is byte-for-byte unchanged.

Button GPIO mapping confirmed against Seeed's Zephyr board definition (`gpio-keys`: refresh=3, left=4, right=5, all active-low).

## Testing

- Verified on reTerminal E1001 hardware: pressing either white button wakes the device and advances the displayed screen.
- Builds cleanly for `seeed_reTerminal_E1001` and `seeed_reTerminal_E1002`.

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -163,6 +163,17 @@ enum WIFI_CONNECT_RETRY_TIME // Time to sleep before trying to connect to the Wi
 #define DEVICE_MODEL "reterminal_e1003"
 #endif
 
+// Optional extra "next" navigation buttons in addition to the PIN_INTERRUPT
+// wake button. The reTerminal E1001/E1002 expose two more front buttons next to
+// the green button: Left (GPIO4) and Right (GPIO5), both active-low. Pressing
+// either wakes the device and triggers a normal refresh, which advances to the
+// next playlist item ("next" screen).
+#if defined(BOARD_SEEED_RETERMINAL_E1001) || defined(BOARD_SEEED_RETERMINAL_E1002)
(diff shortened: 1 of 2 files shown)


tajchert:
I realised this act the same as existing green button (no new functionality) but add some new code so seems unnecessary. 

If somebody needs to utilize those buttons hope this code helps as a starting point (work well).

PR #448, opened 2026-06-29
Fix missing punctuation (apostrophes) in display fonts by expanding ASCII range

taichikuji:
**Description:**
**What is the problem?**
On certain screens (like the WiFi error screen), text containing apostrophes was rendering incorrectly with a blank space instead of the character (e.g., `"Can't establish WiFi connection."` rendered as `"Can t"`). 

**What is the root cause?**
The embedded `BB_FONT` headers used by the display (`nicoclean_8.h` and `Inter_18.h`) were missing the apostrophe glyph (ASCII 39) because they were originally generated with a restricted ASCII range. 

**What does this PR do?**
This PR regenerates the two font headers from their source `.ttf` files using the `bb_epaper` font conversion tool, ensuring the full standard printable ASCII range is included.

**Specific Changes:**
* Regenerated `include/nicoclean_8.h` (size 8) from NicoClean TTF.
* Regenerated `include/Inter_18.h` (size 18) from Inter TTF.
* Expanded the glyph export range to `first: 32, last: 126` to explicitly include all standard punctuation (including `'`, `!`, `?`, etc.).

**Verification:**
* [x] Replaced the existing headers with the newly generated ones.
* [x] Verified header metadata confirms the new range (`first: 32, last: 126`).
* [x] Successfully compiled the firmware using `pio run -e trmnl` with no errors.


taichikuji:
Do not accept the PR yet, it reduces the Point Size, and I believe this would cause issues rather than help. I am revising this again.

taichikuji:
Closing this out until I figure it out or someone patches it first.

Issue #456, opened 2026-06-30
Excessive battery drain on OG on v1.8.6


schrockwell:
Support has notified us that some users are experiencing excessive battery drain on OG devices on v1.8.6 - battery life only lasting a week! :(

schrockwell:
Per standup discussion:

1. per Larry, this could be explained by excessive Wi-Fi access (bad credentials forcing retries, etc), possibly introduced by #390
2. support is only seeing this issue on older OG devices - sometimes a battery or PCB replacement fixes it?
3. we are going to punt this until we are more certain it is a firmware issue

schrockwell:
This is explainable by two other issues. Closing this one.

1. bug in TRMNL core (fixed today)
2. #480

PR #457, opened 2026-06-30
On OG, don't set PIN_INTERRUPT to INPUT

schrockwell:
Addresses #456.

Commit 36978872ec02d6d085c60285ef0a004602ab849b added OG Gen2 support, but it also affected the OG in an unexpected way.

By setting `PIN_INTERRUPT` to the `INPUT` mode, that leaves it floating, which means wake-ups could be triggered randomly and excessively when the device is supposed to be in deep sleep.


schrockwell:
The PCB has an external pull-up resistor on this line, so this analysis is not correct and isn't the culprit for the battery drain. Closing.

Issue #334, opened 2026-04-18
E1002 refresh inconsistencies


benjaminfenech:
Hi all,

I recently purchased a seeed studio E1002 so I could flash DIY TRMNL. 

Seems to have flashed correctly and working as intended, but I have an issue with how the screen refreshes when deploying a playlist. When the screen refreshes with the TRMNL logo in the bottom right, it seems to refresh correctly and display an entire white screen with a black logo in the bottom right.

When deploying a playlist, in my case the TRMNL maintained weather plugin, the screen refreshes and and ends with some blue flashes which causes a blue vignette around the border which fades in to white towards the middle. I took a video but the whole screen just looks blue in the recording:

Album of two side by side here in case embeds don't load well https://imgur.com/a/nGytynl

This is currently firmware 1.7.6. From the looks of it there have been some changes to how refreshing works between firmware versions, but this is the first time testing this device. Is there anyway to change how the refresh cycles between colours?

DavidOrtmann:
I've seen a similar issue when users didn't flash the correct firmware by accident. Are you sure you've selected the e1002 firmware when visiting www.trmnl.com/flash? I have you myself and it's working great.

benjaminfenech:
It's definitely a possibility, I will try again and make sure E1002 is selected

BoxenOfDonuts:
Having the same issue on my E1002 as well. Rolling back to the 1.6 firmware fixes it. Both 1.7.4 and 1.7.6 firmware have the blue tint

ryanckulp:
closing as we think this has been resolved, as reported in #308 and #372.

Issue #372, opened 2026-05-28
E1002 Issues on 1.7.x Firmware


BoxenOfDonuts:
I have been having a few issues with the 1.7.x versions.

When flashing 1.7.4 or 1.7.6, using the web flasher, the wifi startup page never shows, it is just a white screen with the trmnl logo. When flashing 1.6.7 it displays correctly.
```sh
[20:57:27]I: src/display.cpp [1406]: display_show_image end
[20:57:28]I: src/bl.cpp [257]: Display TRMNL logo end
[20:57:28]I: src/bl.cpp [261]: Firmware version 1.7.6
[20:57:28]I: src/bl.cpp [262]: Arduino version 2.0.17
[20:57:28]I: src/bl.cpp [263]: ESP-IDF version 4.4.7
[20:57:28]I: src/filesystem.cpp [206]: Filesystem Usage: 0/113201
[20:57:28]I: src/bl.cpp [2366]: NVS Usage: 77/630 entries (12.22
[20:57:28]I: src/bl.cpp [316]: WiFi NOT saved
[20:57:28]I: src/bl.cpp [318]: FW version 1.7.6
[20:57:28]I: src/bl.cpp [2191]: src/bl.cpp [2191]: esp flash size: 33554432
[20:57:28]
[20:57:28]I: src/display.cpp [1882]: Free heap in display_show_msg - 262132
[20:57:28]I: src/display.cpp [1883]: maximum_compatibility = 1
[20:57:28]
[20:57:28]I: src/display.cpp [1886]: Free heap after bbep.allocBuffer() - 262132
[20:57:28]I: src/display.cpp [1891]: Display set to white
[20:58:03]I: src/display.cpp [1911]: display_show_msg2 start
[20:58:03]I: src/display.cpp [1966]: wifi connect case
[20:58:03]I: src/display.cpp [2006]: Start drawing...
[20:58:37]I: src/display.cpp [2014]: display_show_msg2 end
```

It also has this blue hue when displaying plugins.

<img width="4032" height="3024" alt="Image" src="https://github.com/user-attachments/assets/d54bf167-6e29-4cb5-8497-8f2829b94c71" />

[...]

PR #458, opened 2026-06-30
Fix E1002 Spectra6 update transport

brettinternet:
This change fixes a E1002 color update issue by bypassing the bb_epaper Arduino-SPI write/refresh path for E1002 and using a direct ESP-IDF SPI Spectra6 update path instead. I tried to keep non-E1002 refresh behavior unchanged. I used AI assistance with this work. #334 

Tested on my affected E1002 hardware.

Before:

before

After:

after


dmellok:
Tested #458 on a reTerminal E1002 that was hard-bootlooping on `main`/v1.8.8 — `rst:0x10 (RTCWDT_RTC_RST)` repeating every few seconds, screen flashing, never finishing a paint (console confirmed via UART0). No amount of power-cycling helped, and it looped even when powered independently of the host (ruling out DTR/RTS auto-reset).

After building `seeed_reTerminal_E1002` from this branch and flashing, it boots once (`rst:0x1 (POWERON)`) and paints cleanly — no reboot loop, no watchdog resets. So this fixes a second E1002 panel beyond the author's, for what looked like the severe end of the same Spectra6 update-path incompatibility. Thanks for the fix! 🙏

brettinternet:
That's great! Thank you for testing!

brettinternet:
I think the solution here might be a lighter touch compared to my solution in this PR: https://github.com/usetrmnl/trmnl-firmware/issues/372#issuecomment-4866772408

ryanckulp:
@schrockwell just tested this, LGTM. (made a new swatch throwaway plugin)

e1002-pr-458-swatch

schrockwell:
@bitbank2 will have to review this for correctness. He might be interested in pulling these changes upstream, directly into bb_epaper. Then everybody using the library would get the benefit.

If he approves this, we could merge it in "for now" to move things along. Later, we can update the bb_epaper dependency after it gets the fix.

schrockwell:
We are trying to avoid bloating already-large files like `bl.cpp ` and `display.cpp`, so I extracted the new `spectra6_*` functions into `displays/spectra6.cpp`.

I also merged in `main` and resolved conflicts to the best of my ability.

@brettinternet please give this another test. I don't have the hardware.

ryanckulp:
FWIW i just built/uploaded this updated branch and it behaves as well as before.

brettinternet:
Thank you for the changes and merge 🙏

bitbank2:
FYI - this change will be removed in a new firmware version since the Spectra6 code was improved in bb_epaper.

PR #467, opened 2026-07-03
display SSID on WiFi Connect screen

loucyan:
- when setting up multiple devices it's not clear which TRMNL is which
- this makes it more clear which wifi network is actually in question

we made this using Claude with @niklasbabel so it's only a proof of concept

IMG_2157

### DIFF
--- a/lib/wificaptive/src/WifiCaptive.cpp
+++ b/lib/wificaptive/src/WifiCaptive.cpp
@@ -22,6 +22,15 @@ void WifiCaptive::setUpDNSServer(DNSServer &dnsServer, const IPAddress &localIP)
     dnsServer.start(53, "*", localIP);
 }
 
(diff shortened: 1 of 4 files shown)


schrockwell:
Good stuff. I solved a merged conflict and did a bit of clean-up. Thanks!

schrockwell (approved):

schrockwell:
We have an issue with this on OG. The font is different than X, and the single quote marks around the SSID don't show up. Are they missing from the font? Did we use the wrong quotes? Should we switch to double-quotes, or something else?

IMG_1507

schrockwell:
nvm, fixed with double quotes.

loucyan:
> nvm, fixed with double quotes.

@schrockwell interesting, I couldn't get it to work with double quotes 🫠

PR #468, opened 2026-07-03
Escape commas in ESP-AT string parameters

kory-smith:
Hey, guys. This was Claude-generated, so if it's trash, please let me know. But the problem is real. If you try to enter a password (5G modem only) that contains a comma, the password flow breaks, because the comma is interpreted as a delimiter.

From Claude:

Per the ESP-AT command reference, backslash, double-quote and comma are special characters in string parameters and must be backslash-escaped. Modem::connectToNetwork() only escaped backslash and quote, so an SSID or password containing a comma broke AT+CWJAP parameter parsing and the 5 GHz join on TRMNL X failed with a generic WiFi error.

Add an escapeAtParam() helper, use it for AT+CWJAP ssid/password and for inline AT+HTTPCLIENT URLs (AT+HTTPURLCFG keeps the raw URL since it takes length-prefixed bytes). getSignalRssi() and parseNetworks() need no change: they skip the quoted fields by quote position before splitting on commas.


schrockwell:
Thanks @kory-smith. I consolidated this into the existing `string_utils.cpp` and renamed it to `escape_modem_param()` because the acronym "AT" is easy to confuse with the word "at".

PR #470, opened 2026-07-04
Fix WiFi credential pair last-used tracking

### LINKED ISSUE #285: bug - remembering multiple WiFi networks

silasbrookshaha:
Fixes #285.

## Summary
- Track the last-used WiFi entry by the full saved credential pair instead of SSID only.
- Avoid skipping or marking the wrong saved entry when the same SSID has multiple stored credential variants.
- Close the Preferences handle after updating the last-used index.

## Verification
- git diff --check

Note: PlatformIO is not installed in this environment, so I could not run the firmware unit or hardware tests locally.


Omeryl:
Automatic AI submissions that do not test the build are a bit odd to me.

silasbrookshaha:
Thanks, that's fair feedback. I installed PlatformIO in a clean temporary environment and ran the available local checks.

Verification now completed:
- `pio test -e native` passed: 52/52 test cases succeeded.
- `pio test -e trmnl_test --without-uploading` successfully built the TRMNL integration-test firmware and reached the hardware serial-test wait.

I still do not have physical TRMNL hardware here, so I could not run the on-device integration tests. I also tried `pio run -e trmnl`, but this local toolchain hits a PlatformIO/ESP-IDF duplicate `esp_efuse_fields.c.o` target conflict before compiling project code.

schrockwell:
This is a nice fix. Smoke-tested on the OG. Also reformatted per the new `clang-format` rules introduced on `main`.

Thanks for your contribution!

PR #479, opened 2026-07-07
Fix sleep issue when playlist has one item

### LINKED ISSUE #469: Device stays in sleep mode forever if there's only one playlist item and it's stale
There is a smattering of users who are experiencing odd sleep behavior. I am one of those users. Since I have so many TRMNL devices, many of them just have one playlist item. The bug is that if a device has only one playlist item and the data for that playlist item doesn't change very often. The TRMNL will never leave the sleep screen, presumably because it decides it doesn't need to refresh because of stale data.
 The bug goes as follows:

For example, my TRMNL on my desk displays a to-do list connected to Todoist. If between sleep and wake up none of my tasks have changed, the device will stay asleep until I click the button on the back or I change something in Todoist. I investigated this somewhat with Claude, and I'm almost positive it is a bug with screen staleless and wake-up logic, I will post Claude's summary below:

> Device stays on sleep screen after wake — szPrevFile not updated on the download path
> 
> Affected: v1.8.1+ (introduced by 1b6dd33; present in v1.8.8). Symptom: a device using sleep mode whose playlist serves an unchanged image at wake never repaints — the sleep image stays until the plugin's data changes. Most visible with single-item playlists; also reachable with multi-item playlists when schedules/TTL make the resolver re-serve the last-displayed item.
> 
> Mechanism: 1b6dd33 added RTC_DATA_ATTR char szPrevFile[36] and a skip in downloadAndShow(): when the served filename is already in SPIFFS and equals szPrevFile, skip the refresh ("The image hasn't changed since the last wakeup"). But strcpy(szPrevFile, ...) only happens in that cached path. The fresh-download path (bl.cpp ~1969 in v1.8.8 — writes the NVS path keys) draws the image without updating szPrevFile. The sleep screen is nearly always a fresh download (see secondary bug), so: bedtime draws sleep, note still holds the playlist filename; wake serves that same filename (server only rotates names when content changes); note matches; no repaint. szPrevFile survives deep sleep (RTC), so the panel shows the sleep image indefinitely.
> 
> Fix: update szPrevFile wherever an API image is drawn — i.e. add the strcpy to the download path (and any other display_show_image call that changes the glass). Invariant: szPrevFile must always equal what's physically displayed.
> 
> Secondary bug, same area: filesystem_purge_old_file() computes file age via atoi(&s[strlen(s)-10]) — an out-of-bounds read for names shorter than 10 chars ("sleep"), which typically parses as 0 and deletes the cached sleep image on every purge. This is what forces the sleep screen through the download path each night.
> 
> Server context: screen filenames are <blob>-<updated_at> and deliberately stable when plugin data is unchanged; that contract predates this firmware change by a year.
> 
>

schrockwell:
Fixes #469 

## The problems

1. we were not setting `szPrevFile` in all situations where an image was displayed
2. we weren't accounting for filenames that don't have a timestamp, like "sleep.png"

## The solution

1. refactor: extract "previous image" logic into new `DisplayedImage` namespace
2. fix: add `DisplayedImage::remember()` calls where they were missing
3. fix: improve filename timestamp parsing

### DIFF
--- a/include/bl.h
+++ b/include/bl.h
@@ -30,4 +30,6 @@ void logWithAction(LogAction action, LogLevel level, const char *message, time_t
 bool submitLogString(const char *log_buffer);
 bool storeLogString(const char *log_buffer);
 
+void fixFileName(const char *src, char *dest);
+
 #endif
--- a/include/filesystem.h
+++ b/include/filesystem.h
@@ -87,4 +87,11 @@ bool filesystem_file_delete(const char *name);
  */
 bool filesystem_file_rename(const char *old_name, const char *new_name);
 
+/**
+ * @brief Function to extract the timestamp from a filename
+ * @param filename filename
+ * @return timestamp as uint32_t (0 if timestamp suffix is not present or invalid)
+ */
+uint32_t filesystem_extract_timestamp(const char *filename);
+
 void list_files();
--- a/src/bl.cpp
+++ b/src/bl.cpp
@@ -383,6 +383,7 @@ static void showLastImageAndSleep()
     if (buf && file_size > 0) {
       display_show_image(buf, file_size, true);
       free(buf);
+      DisplayedImage::remember(curPath.c_str());
     }
   }
   goToSleep();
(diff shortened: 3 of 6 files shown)


schrockwell:
I sent @kory-smith bootleg firmware and he has confirmed the fix on his OG.

PR #488, opened 2026-07-09
Fixed AP list refresh on X, added proper indication for rescan

GoReNeY:


schrockwell:
Tested on X, working as expected.

PR #495, opened 2026-07-14
Fixed OG partial update to use previous image in EPD RAM to keep the balance of charge

bitbank2:
Previous versions of the trmnl firmware did unbalanced pushes to get partial (non flickering) updates. This PR changes the logic to make use of the EPD's low power RAM to hold the previous image and do normal partial updates which respect the balance of charge (push equally on white and black particles). The EPD sleep mode was changed from 'deep' to 'light' to preserve the RAM for a cost of about 4uA in current (insignificant). The loading screen has switched to a FAST update instead of partial and partial updates will only occur when transitioning from a 1-bit image to a new 1-bit image.

### DIFF
--- a/src/bl.cpp
+++ b/src/bl.cpp
@@ -1700,7 +1700,7 @@ void load_prev_image(void)
   size_t content_size = 0;
   if (content_size > 0) {
     // Decode it into the previous buffer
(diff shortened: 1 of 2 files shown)


schrockwell:
Tested and working as expected - thanks. Will merge this.

I'd be curious to see if ghosting can be reduced in a future PR - you mentioned there were some tricks we could play.

PR #508, opened 2026-07-23
Refactoring touchbar logic, added progress for touchbar indicator

GoReNeY:


schrockwell:
@GoReNeY I merged in from `main`, resolved the conflicts, and reformatted.

I am not seeing the progress bar when refreshing. I only see the white fill on the indicator. I tested it before and after merging in `main` and I see no difference, so I don't think it's a problem introduced by the merge. I am using the Alias plugin but I have caching disabled, so I don't think the image is loading from the device cache.

https://github.com/user-attachments/assets/ea105897-7358-4681-8a6e-b8207396f76e

GoReNeY:
@schrockwell the progress is only shown for HOLD gestures, on the video you're using TAP gestures

schrockwell:
Oh, I misunderstood. I was expecting a progress bar to show image download progress. Disregard.

schrockwell:
In #439 (see https://github.com/usetrmnl/trmnl-firmware/pull/439/changes/de0b614ea4bc7baee39c88160e33d2cbe5245583) we made the loading screen fade the current image lightly into the background behind the logo. Is it possible to keep that with this change? Or do we need to do a full screen clear?

GoReNeY:
Well I don't see any reasons why we shoudn't keep that change, since progress bar is drawn using partial refresh

schrockwell:
Hi @GoReNeY - is this still in progress, or can I merge it in? I can look into the partial refresh as well.

GoReNeY:
@schrockwell Yeah I guess it can be merged, on my devices no issues so far, though that changes should to be tested properly before pushing to release

schrockwell:
Merged in main. Tested left/right/center taps, swipes, Wi-Fi reset. 

The partial-refresh loading screen came in from main, so no extra work there.

PR #511, opened 2026-07-24
Improve X waveform for 4BPP images + `FastEPD` refactor

ronakjain90:
This PR Improves 4bpp grayscale on TRMNL X with 38-pass graymatrix (supplied by manufacturer) and refactors the X-class display code out of `display.cpp` behind a small interface.

Changelog: 
  - Refactor: move the fastepd code into it's own namespace.
  - TRMNL X: 38-pass gray matrix
  - init panel at 40 MHz
 
Tests:
  - [X] TRMNL X
  
Result:
Before:
620359c7-ea70-42be-81fe-3778d2900637 png
After
WhatsApp Image 2026-07-24 at 18 16 03

PR #512, opened 2026-07-24
Switches the gray matrix from 9 to 38 steps to improved gray photo rendering

bitbank2:
This change switches the current 9-step gray matrix for a 38-step matrix. The new matrix takes a bit longer to render, but gives better results for dithered photos. The pure gray curve on "chart graphics" is a bit worse, but is less noticeable than the artifacts which appeared on photos with the older matrix.


schrockwell:
@bitbank2 Does this PR supersede #511?

bitbank2:
> @bitbank2 Does this PR supersede #511?

Ronak and I are still discussing this. PR 511 is dead. I think this gray matrix should be utilized only for photos on the X and the original 9-step matrix should be used for non-photo images.

PR #519, opened 2026-07-27
Add 5 GHz support to Gen2

schrockwell:
This PR supersedes #407. After that PR was opened, we separately added the `is5GHz` flag to WiFi credentials in #463, which made most of the previous PR moot.

This new PR is the new, minimal changeset to enable 5 GHz connections on the OG Gen2.

- [x] test on dual-band network, 2.4 GHz
- [x] test on dual-band network, 5 GHz

### DIFF
--- a/lib/wificaptive/src/connect.cpp
+++ b/lib/wificaptive/src/connect.cpp
@@ -127,6 +127,18 @@ void disableWpa2Enterprise() {
   esp_wifi_sta_wpa2_ent_clear_ca_cert();
 }
 
+static void setWiFiBand(const WifiCredentials &credentials) {
+#ifdef CONFIG_SOC_WIFI_SUPPORT_5G
+  if (credentials.is5GHz) {
+    Log_info("Configuring for 5GHz only");
+    esp_wifi_set_band_mode(WIFI_BAND_MODE_5G_ONLY);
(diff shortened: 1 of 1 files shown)


lib/wificaptive/src/connect.cpp
```diff
@@ -127,6 +127,18 @@ void disableWpa2Enterprise() {
...
 
+static void setWiFiBand(const WifiCredentials &credentials) {
+#ifdef CONFIG_SOC_WIFI_SUPPORT_5G
+  if (credentials.is5GHz) {
+    Log_info("Configuring for 5GHz only");
+    esp_wifi_set_band_mode(WIFI_BAND_MODE_5G_ONLY);
+  } else {
+    Log_info("Configuring for 2.4GHz only");
+    esp_wifi_set_band_mode(WIFI_BAND_MODE_2G_ONLY);
+  }
```
schrockwell, line 138: @bitbank2 I added this conditional branch to ensure that the user's 2.4 GHz network selection is honored. Otherwise, the radio would remain in auto mode, and it might pick the 5 GHz network instead. Please let me know if this isn't desirable.

bitbank2:
Looks correct - thanks :)

PR #523, opened 2026-07-29
Add ESP32-C5 N8R8 for OG Gen2

schrockwell:
The new Gen2 V2 boards were changed from N16R8 to N8R8, so I've added the new board config and partition table, and changed from 16MB to 8MB where relevant.

I also moved the partition CSVs into `boards/` to clean up the project root.


bitbank2:
Looks good to me. My only thought is that the main program and OTA partitions don't need to be 3MB each. Our code is currently about 1.3MB, so setting them each to 2MB and adding the extra 2MB to the SPIFFS partition might be helpful for future cached file storage.

schrockwell:
@bitbank2 Good idea. The gen2 build is actually up to 1.7 MB, so I opted for 2.5 MB app partitions for a bit more headroom.

```
RAM:   [==        ]  18.3% (used 59876 bytes from 327680 bytes)
Flash: [=======   ]  68.5% (used 1796860 bytes from 2621440 bytes)
```

PR #526, opened 2026-07-30
Add single and triple button feedback beeps

joelhawksley:
## Summary

- beep once when a button press is released before the long-hold threshold
- retain the existing two-beep feedback at the WiFi reset threshold
- beep three times when the soft-reset threshold is reached
- use the refactored `BaseBuzzer` interface introduced after #402

## Test plan

- `pio run -e seeed_reTerminal_E1001`


schrockwell:
Closing as a dupe of #527

PR #527, opened 2026-07-30
Add single and triple button feedback beeps

joelhawksley:
## Summary

- beep once for short presses, double clicks, and medium holds after the action is classified
- acknowledge quick GPIO-wakeup presses even when the button is already released before parsing starts
- retain the existing two-beep feedback at the WiFi reset threshold
- beep three times when the soft-reset threshold is reached
- use the refactored `BaseBuzzer` interface introduced after #402

## Test plan

- `pio run -e seeed_reTerminal_E1001`
- `PLATFORMIO_CORE_DIR="$HOME/.platformio-x" pio run -e TRMNL_X_E1003`

### DIFF
--- a/src/button.cpp
+++ b/src/button.cpp
@@ -16,6 +16,9 @@ static unsigned long wait_for_button_release(unsigned long start_time) {
     }
     delay(10);
   }
+  if (millis() - start_time >= BUTTON_SOFT_RESET_TIME) {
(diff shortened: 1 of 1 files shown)


joelhawksley:
@schrockwell thanks for landing #402! This PR adds beeps for the other button press events.

schrockwell:
Easy enough, LGTM

PR #532, opened 2026-08-01
Snake game

ryanckulp:
testing partial refresh, playable on X. just hold middle of touchbar for 5 seconds.

https://github.com/user-attachments/assets/a4b873ba-9b87-4bc3-a887-4e33fbf6788e

### DIFF
--- a/include/snake.h
+++ b/include/snake.h
@@ -1,0 +1,53 @@
+/******************************************************************************
+ * @file        snake.h
+ * @brief       Easter egg: Snake, played on the touchbar.
+ *
+ *              Entry: hold the middle of the touchbar for 5 s. Left tap
+ *              turns the snake 90° counter-clockwise, right tap turns
+ *              clockwise.
+ ******************************************************************************/
+
+#ifndef SNAKE_H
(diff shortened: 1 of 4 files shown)


bitbank2:
Cute. It might be easier for the user to control with a menu at the bottom of the display showing 4 regions of the touchbar signifying 4 direction changes.
The update rate could improve if the eink power is left on between steps. For some reason the LLM generated code to turn it off after each update.

I'm okay with adding it as-is too.

ryanckulp:
i prefer as-is because we don't currently have 4x zones dialed in, and folks will learn quickly enough the clockwise/counter-clockwise paradigm. while playing myself the battery seemed to drain quite a bit so, also OK with keeping updates off.

bitbank2:
> i prefer as-is because we don't currently have 4x zones dialed in, and folks will learn quickly enough the clockwise/counter-clockwise paradigm. while playing myself the battery seemed to drain quite a bit so, also OK with keeping updates off.

The change won't affect the battery, it will just remove wasted time between updates.

Issue #505, opened 2026-07-21
Add special function value in API Display response header


bkuhlmann:
## Why

Hey, so I'm working on Terminus Issue [345](https://github.com/usetrmnl/terminus/issues/345) which is asking for special function support. The problem is that our firmware is not capturing the value of the special function when sent to the device. Example (Display API response):

``` json
# Truncated for brevity
{
  "special_function": "add_wifi"
}
```

I've tried force pressing the button on the back of my OG device (and letting the device wake up to accept the command) only to find it never stores this information. I was able to confirm by killing my server, forcing the device to refresh (which would fail) and then looking at the corresponding device log, once the server came back up *and* device sent over the error logs, only to find that nothing is recorded for the _Special Function_:

<details>
  <summary>Device Log Screenshot</summary>

<img width="1338" height="900" alt="Image" src="https://github.com/user-attachments/assets/4e5485bf-a8f4-4aa4-8102-cd66c94a931f" />
</details>

## Steps to Reproduce

1. Send any special command to a device and notice that it never stores this information via the API Log POST request or via the Display API GET request.

## How

1. Clarify terminology by calling these "commands" that is what Terminus uses but I think there is a proposal within Core to call these "controls". Either way, being consistent with terminology would be better than calling these "special functions".
1. Add a HTTP Header (like: `command`) which includes the value of the command. For example, if the user has their command button configured to "pause" then the API Display request should include `command=pause` in the header. This allows the server to know the device got the command and now the server can process that command dynamically by interrupting the current playlist by *not* sending the next item in the playlist but, instead, rendering the appropriate screen for that command instead.

## Notes

- **Firmware**: 1.8.10

I can work around most of the special commands by handling everything server-side but with "Add WiFi" support, I actually need a way for the device to trigger the captive portal. That's the only special command that requires the device to listen and do as requested.

bkuhlmann:
**Update/Correction**: I was able to confirm that the `add_wifi` command was honored by the device but I'm still not seeing this value be recorded in the device logs. In fact, it seems like having an *info* log to acknowledge use of the command would be helpful. This would add an extra HTTP request but only for when folks use the button to issue a command (which would be infrequent at least).

ryanckulp:
looks like a good start is to change this:
https://github.com/usetrmnl/trmnl-firmware/blob/9a1ab1195a732ac9a965f8a91fe15b81d777b23c/lib/trmnl/src/api-client/request_headers.cpp#L39

[...]

PR #537, opened 2026-08-04
Screen Wiper

ryanckulp:


schrockwell:
Tested on my X. I had some very minor ghosting because I only rotate between two plugins on my X, and this absolutely reduced it.

After the wiper runs, the display just sits on a white screen until the next scheduled refresh. Should we do something nicer here? "Screen squeegee complete!" message? Refresh to load the next playlist item (would get caught in a loop if there is only one playlist item)? Something else?

schrockwell:
One idea

image

ryanckulp:
i vote we skip to the next playlist item automatically. i think some folks will keep this in their playlist indefinitely for daily artifact removals. i don't want to give them work or block their device functionality. 

but re your demo, perhaps the end says "Screen wipe complete, resuming service." or something like that, for 1-2 seconds, then skips to the next item. that way people don't get confused, _what is this_?

schrockwell:
> i vote we skip to the next playlist item automatically. 

Duh, this makes the most sense. Thanks, will implement.

> i think some folks will keep this in their playlist indefinitely

The way I imagine this, a user would want to wipe 1x per day. Making that part of the playlist is hacky, and would be a pain to configure so that it runs exactly once (would need to make it Important, set a schedule, and ensure the playlist items' refresh is set exactly right).

Instead of making this a plugin, I'd like to propose a new device setting toggle like "automatically wipe screen" which would either (a) wipe the screen immediately before sleep, if sleep is enabled, or (b) wipe it at 3am local time.

Or, the user could trigger a one-time "wipe screen on next refresh" which would return to the normal playlist afterwards.

bkuhlmann:
Related to [Issue 505](https://github.com/usetrmnl/trmnl-firmware/issues/505), this would fold in nicely if the firmware could send back the command used to the server via the API Display GET request. In short, pressing the device's button for ~2 seconds to invoke the special function (i.e. command), would allow any server (Core or BYOS) to interrupt the playlist with the wipe. The wipe then becomes a programmable command in a list of infinite commands a user could program their device to use.

ryanckulp:
> Instead of making this a plugin, I'd like to propose a new device setting toggle like "automatically wipe screen" which would either (a) wipe the screen immediately before sleep, if sleep is enabled, or (b) wipe it at 3am local time.

this is fine, but let's ship as-is in plugin format to give a few users a go at it, see if it solves their problem, then promote to a new setting. the plugin route is a nice v1 because it's only a few LOC.

schrockwell:
Tested on OG, X, and BWRY.

Works great on OG and X.

On BWRY, a single wipe takes 15 seconds. So the prescribed 60 wipes would take 15 minutes to complete. I am going to lower the BWRY to 1 wipe instead, since it likely will not have any ghosting anyway.

ucffool:
> Making that part of the playlist is hacky, and would be a pain to configure so that it runs exactly once (would need to make it Important, set a schedule, and ensure the playlist items' refresh is set exactly right).

If your playlist is naturally 15 minute refresh, you pick a 15m window, set important, set schedule, and make screen wiper Duration 30m (makes sure it doesn't trigger twice). Seems simple enough to setup end of day.

PR #540, opened 2026-08-06
Gas gauge fixes

GoReNeY:
- Added needed read delays to fix issue with wrong design scale
- Reorganized gas gauge related logic and moved to appropriate files
- Added check if chip havsappropriate golden file (code checks design scale on connection, where 1 is recognized as 1cell config and 10 as 2cell config)
- Added values range checks and gas gauge reset if data is invalid

### DIFF
--- a/include/battery.h
+++ b/include/battery.h
@@ -49,12 +49,40 @@ private:
 };
 
 #ifdef INCLUDE_BQ27427
-/// @brief Reads the battery voltage from the BQ27427 fuel gauge.
+/// @brief Reads the battery voltage from the BQ27427 fuel gauge, and gathers
+///        the full set of gas gauge readings (SOC, health, current, etc.).
 class BQ27427Battery : public BaseBattery {
 public:
-  float readVoltage() override;
+  /// @brief Battery voltage from the last gaugeInit() snapshot.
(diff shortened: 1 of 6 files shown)


schrockwell:
Love the refactor. Please also see my findings in https://github.com/usetrmnl/core/issues/4116 about the wrong sign for current sensing.

schrockwell:
I tested this on my X (two batteries) and saw no change to the behavior. After refreshing 14 times on battery, it reported the remaining capacity as increasing from 11,210 to 11,240 mAh. I also noted the current measurement was still positive (wrong).

Then, I adjusted the `CC Cal` register value from `0xAD` to `0x2D`, which flips the sign of the `CC Gain` parameter. It's not documented in the datasheet or the reference manual anywhere. See https://github.com/usetrmnl/core/issues/4116#issuecomment-5208618358 for threads on the TI forum. One person claims it was programmed wrong from the TI factory. Seems possible.

After 10 refreshes with the new `CC Cal` value, negative current was reported during battery drain, and the capacity dropped as expected, from 11,250 to 11,220 mAh.

So, while this PR is a good improvement, it doesn't fix the issue. I will apply the current-sensing fix in a separate PR so that we can test each fix separately.

PR #544, opened 2026-08-07
Add user-triggered shipping mode for reTerminal E1001/E1003

joelhawksley:
## Summary

Adds shipping mode for the Seeed reTerminal E1001/E1003 activated by holding the button for 30s and indicated by a series of four beeps.

These boards use the SY6974B I2C charger PMIC, so shipping mode reads its status register to detect whether USB/charger power is present.

## Staged button-feedback beeps

Extends the existing `buzzer()` interface with staged feedback during a hold, so the user hears each threshold:

| Hold | Beeps | Action |
|------|-------|--------|
| n/a | 1 | refresh |
| 5 s | 2 | long press |
| 15 s | 3 | soft reset |
| 30 s | 4 | shipping mode |

The beeps are a no-op on boards without a buzzer (base `BaseBuzzer`).

## Scope / notes

- Reuses the existing shipment helpers (`qa.cpp`) and `READY_TO_SHIP` / `SHIPPING_MODE` screens already in the tree.
- Gated by a new `SHIP_MODE_SUPPORTED` define (set only for E1001/E1003, alongside the `CHARGER_*` I2C config).
- No behavior change for boards without `SHIP_MODE_SUPPORTED`.

## Testing

Builds succeed for:

- `seeed_reTerminal_E1001`
- `TRMNL_X_E1003`
- `trmnl` (OG, to confirm the shared `button.cpp` change is a no-op regression-wise)

### DIFF
--- a/lib/wificaptive/src/connect.h
+++ b/lib/wificaptive/src/connect.h
@@ -1,10 +1,15 @@
 #pragma once
 
 #include <WiFiType.h>
-
+#include <functional>
 #include "wifi-helpers.h"
 #include "wifi-types.h"
 
 WifiConnectionResult initiateConnectionAndWaitForOutcome(const WifiCredentials credentials);
 wl_status_t waitForConnectResult(uint32_t timeout);
 void disableWpa2Enterprise();
+
+// Registers a callback that is invoked repeatedly while waiting for a WiFi
+// connection to complete. Lets the caller poll hardware (e.g. the button) during
+// the otherwise blocking connect attempt. Pass nullptr to clear.
+void setConnectTickCallback(std::function<void()> cb);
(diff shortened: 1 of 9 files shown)


lib/wificaptive/src/connect.h
```diff
@@ -1,10 +1,16 @@
...
 #include "wifi-types.h"
 
 WifiConnectionResult initiateConnectionAndWaitForOutcome(const WifiCredentials credentials);
 wl_status_t waitForConnectResult(uint32_t timeout);
 void disableWpa2Enterprise();
+
+// Registers a callback that is invoked repeatedly while waiting for a WiFi
+// connection to complete. Lets the caller poll hardware (e.g. the button) during
+// the otherwise blocking connect attempt. Pass nullptr to clear.
+void setConnectTickCallback(std::function<void()> cb);
```
joelhawksley, line 16: (This allows the button presses to work during wifi connection)

PR #545, opened 2026-08-07
Move user-facing strings and QR URLs into config.yml

joelhawksley:
## Summary

Moves the firmware's user-facing display text and the on-screen QR code URLs into a single `config.yml`, generated into headers at build time. Defaults reproduce the stock TRMNL strings and URLs exactly, so an unmodified `config.yml` builds identical firmware.

This makes the firmware straightforward to localize the firmware without editing C++, such as for device deployments using Terminus, etc.

## How it works

`scripts/generate_branding.py` runs as a PlatformIO pre-build script and emits:

- `include/branding.h` — string / URL `#define` macros.
- `src/wifi_connect_qr.h`, `src/wifi_failed_qr.h` — 66×66 G5 QR bitmaps (same format `bbep.loadG5Image` already consumes, so no display code changes).

## Notes

- `include/branding.h` and the QR headers are generated; added to `.clang-format-ignore` alongside the existing generated headers.
- Captive-portal web page text is intentionally out of scope here for now.
- Wired the pre-build script into the shipping envs (`trmnl`, `trmnl_4clr`, `TRMNL_X`, `TRMNL_X_E1003`, `seeed_reTerminal_E1001/E1002`, `WAVESHARE_397`).

## Testing

Builds succeed for `trmnl` (OG), `seeed_reTerminal_E1001`, and `TRMNL_X_E1003` (X-class), with the generator running as a pre-build step. The QR bitmaps were verified to round-trip to the encoder's QR matrix. `clang-format` passes.


schrockwell:
@joelhawksley When you get a chance, please rebase/merge from main. Some conflicts were introduced by a major refactor on our end.

PR #546, opened 2026-08-09
Add WPA2-Enterprise support on TRMNL X (2.4 GHz)

SethAngell:
## What Was Changed?

Added branching logic to the TRMNL X connection logic to use the 2.4ghz modem for enterprise networks

## Why Was It Changed?

The firmware for the 5GHZ modem doesn't support WPA2-Enterprise WiFi. Previously, this meant that the Network Search and Captive Portal logic skipped WPA2 networks when setting up a TRMNL X. This change adds branching which will include those enterprise networks in the captive portal selection. When chosen, the X falls back to using ESP32-C3's 2.4ghz modem to connect, rather than the dedicated 5ghz modem. This allows corporate bozos like myself access to the 2.8" additional inches of glorious e-ink.


SethAngell:
@schrockwell Hey! Not sure who would be the best person to request a review from, but I saw you added the label so hoping you could provide some direction. Thanks!

GoReNeY:
Tested on OG and X, all OK

PR #549, opened 2026-08-11
Devices, simplified

schrockwell:
1. refactors device pin and display configurations into a simple lookup table
2. DEV_Config.h is finally gone
3. adds support for Seeed E1004
4. adds support for Seeed Sticky
5. adds support for M5 Paper Color
6. adds support for Gen2 4-color


schrockwell:
Superceded by #550

Issue #491, opened 2026-07-10
Move #ifdefs to code edges; embrace OOP


schrockwell:
We need to take advantage of basic OOP principles to clean up the rats-nest of `#ifdef` statements littered throughout the business logic.

I will flesh out this issue with specifics after exploring the codebase more.

schrockwell:
Here is one idea I had - using object composition to separate out individual device components, and a class inheritance tree so that models can easily derive and override logic as needed.

In this example the global singleton instance of a Device would contain private objects for managing power, display, WiFi, and so on. And it would expose public methods that abstract away the specific functionality.

<img width="1156" height="578" alt="Image" src="https://github.com/user-attachments/assets/253ae71e-d04d-45be-a354-535a5939d8f7" />

As for the particular devices, there could be a `BaseDevice` (blue) class with virtual (or empty) methods, and then a few top-level devices for first-party TRMNL devices (yellow), subclassed by the BYOD hardware (green) that imitates them.

<img width="2118" height="1258" alt="Image" src="https://github.com/user-attachments/assets/5c39734f-f3e3-4176-85d3-27cdf5eacbf4" />

PR #550, opened 2026-08-12
Simplify device definitions

bitbank2:
This PR is a modified version of my devices_simplified branch, successfully merged with the latest main.

The changes can potentially break existing device support, so it is critical to test it on every supported device before merging it into main. I have tested the following devices so far:

M5 Paper Mono
M5 Paper Color
Seeed reTerminal Sticky
Seeed E1004

More soon...


bitbank2:
I just simplified the battery reading logic that @schrockwell wrote recently. I folded all of the OG options into a single class (fake, normal, enable switch).
I also tested the code on the Xiao DIY Kit PCB (ESP32-S3) with the 7.5 BW, 7.5 BWR and 4.26 BW

schrockwell:
The fake battery logic was left in there because PR #377 is unresolved, pending customer response. If we are going to remove the fakery, then we can close #377.

src/misc/sensor/environment_sensor.cpp
```diff
@@ -1,24 +1,26 @@
 #include <misc/sensor.h>
 
-#ifdef INCLUDE_ENV_SENSOR
-
 #include <Arduino.h>
 #include <esp_sleep.h>
 #include <time.h>
 #include <trmnl_log.h>
+#include <config.h>
+extern TRMNL_DEVICE *pDevice;
```
schrockwell: Please do not let globals leak into our new code as dependencies. This is what we are fighting against. We're trying to move away from this abuse of global scope.

Instead, pass it as an argument into the constructor of the `EnvironmentSensor` class.

src/bl.cpp
```diff
@@ -2549,6 +2559,22 @@ static void resetDeviceCredentials(void)
   ESP.restart();
 }
 
+#ifdef BOARD_TRMNL_GEN2
+void start_ulp_program() {
```
schrockwell: Please do not add any new functions to bl.cpp. We should only be making it shorter. Please define a new module with a meaningful name (simply `ulp`?) with this function, in a separate file.

src/pins.cpp
```diff
@@ -1,13 +1,20 @@
...
 
-void pins_init(void) {
-  gpio_config_t io_conf = {};
-  io_conf.intr_type = GPIO_INTR_DISABLE;
-  io_conf.mode = GPIO_MODE_INPUT;
-  io_conf.pin_bit_mask = (1ULL << PIN_INTERRUPT);
-  io_conf.pull_down_en = GPIO_PULLDOWN_DISABLE;
-  io_conf.pull_up_en = GPIO_PULLUP_ENABLE;
-  gpio_config(&io_conf);
+void pins_init(void)
```
schrockwell: Same problem here with global scope… pass in `pDevice` as an argument instead.

src/qa.cpp
```diff
@@ -6,10 +6,11 @@
 
 #include "WifiCaptive.h"
 #include "button.h"
+#include "pins.h"
 #include "config.h"
 #include "logo_medium.h"
 #include "logo_small.h"
-#include "pins.h"
+extern TRMNL_DEVICE *pDevice;
```
schrockwell, line 13: 👀

src/api-client/display.cpp
```diff
@@ -10,6 +9,13 @@
 #include <inttypes.h>
 #include <misc/sensor.h>
 #include <trmnl_log.h>
+extern RTC_DATA_ATTR int iPrevWakeTime; // total wake time of the last cycle (for statistics collection)
+extern RTC_DATA_ATTR bool bUsedCachedImage; // if the last image displayed was read from cache (for statistics collection)
+#ifdef SENSOR_SDA
+extern int lastCO2, lastSCDTemp, lastTemp, lastSCDHumid, lastHumid, lastPressure, lastType, lastTime;
+const char *szDevices[] = {"None", "AHT20", "BMP180", "BME280", "BMP388", "SHT3X", "HDC1080", "HTS221", "MCP9808","BME68x","SHTC3"};
+const char *szMakers[] = {"None", "ASAIR", "Bosch", "Bosch", "Bosch", "Sensirion", "TI", "STMicro","MicroChip","Bosch","Sensirion"};
+#endif // SENSOR_SDA
```
schrockwell, line 21: Looks like a merge mistake? Many of these globals are gone, and also are no longer referenced in this file.

include/config.h
```diff
@@ -45,11 +46,11 @@
 #define WIFI_CONNECTION_RSSI                 (-100)
 
 #define DISPLAY_BMP_IMAGE_SIZE 48062 // in bytes - 62 bytes - header; 48000 bytes - bitmap (480*800 1bpp) / 8
-#define DEFAULT_IMAGE_SIZE     48000
-#if defined(BOARD_X_CLASS)
-#define MAX_IMAGE_SIZE 750000 // Use PSRAM on the ESP32-S3 (all X-class boards have PSRAM)
+#define DEFAULT_IMAGE_SIZE 48000
+#if defined(BOARD_HAS_PSRAM)
```
schrockwell, line 50: +1 on the renaming of this flag, much better!

schrockwell:
I want to once again beg that PRs are smaller in the future. It's not tenable to continue like this. I almost want to reject this until it's split out into more manageable pieces. But I understand that we have to move forward.

This is gigantic and contains many unrelated changes that are not dependent on each other. It makes it extremely difficult to review, increases the chances for merge conflicts, and makes it harder to track down when issues were introduced onto master.

EDIT: Based on this week's meeting, we want to get this into main sooner than later. So I'm happy to forego the requested changes, and can help apply the the architecture tweaks in a future PR.

include/battery.h
```diff
@@ -16,36 +16,14 @@ class BaseBattery {
 
   /// @brief Read the battery voltage.
   /// @return voltage in Volts, or a negative value if the reading failed
+  virtual float readVoltage(TRMNL_DEVICE *pDevice) = 0;
```
schrockwell, line 21: There is no need to add this method override. `pDevice` should be stored as a class instance variable. Then any caller can simply do `battery().readVoltage()` without needing to know about the underlying implementation.

src/misc/sensor/environment_sensor.cpp
schrockwell: In order to accomplish this, we'll probably have to initialize one global like:

```c
BaseSensor *sensor = initSensor(pDevice) // renamed sensor() to initSensor(...)
```

which I probably should have done from the start. And then we can update references from `sensor()` to `sensor`. We can adopt a similar pattern for other objects as well. Eventually, this will no longer be a global, but an instance variable in the particular device class (see diagrams in #491).

bitbank2:
Thanks for the feedback @schrockwell , I agree with most of your suggestions. I'm sorry the change grew to this size. Too much work to abandon it now.
I was testing the ULP code and didn't intend for it to be added to this PR. I disabled it for now.
I'll try to get this all cleaned up today/tomorrow.

src/misc/sensor/environment_sensor.cpp
bitbank2: fixed

src/bl.cpp
bitbank2: removed

src/api-client/display.cpp
bitbank2: fixed

include/battery.h
bitbank2: I either add a new method (e.g. init) or have an override function like this. Somehow the pDevice value needs to get into the class instance variable. I think two ways of instantiating the class is worse than 2 types of readVoltage() methods.

bitbank2:
@schrockwell Git is unable to merge 1 simple change between this branch and main even though I fixed it for a merge. Please check this last update for suitability to merge.

schrockwell:
@bitbank2 Thanks for addressing my feedback. I merged in main and resolved the conflicts. Will get this into main next week so that we can continue testing.

This will result in merge conflicts on 13 of the 31 open PRs, so we can work through rebasing and resolving those next.

PR #551, opened 2026-08-13
Fix devices getting stuck offline after a WiFi failure

ikraamg:
Devices can get stuck offline after a WiFi failure. The fallback scan after a failed fast connect returns immediately instead of scanning, and after four failed wakes the device only wakes again on a button press — two devices in one office lost 113 and 67 hours that way.

- fixes waitForConnectResult returning the previous attempt's status, which aborted the fallback channel scan
- continues the retry ladder hourly for a week before it stops waking on a timer
- draws the retry message once instead of on every hourly wake
- logs the 802.11 disconnect reason at ERROR so it reaches the server, instead of only "WL Status: 4" which covers AUTH_FAIL, ASSOC_FAIL, ASSOC_TOOMANY and a handshake timeout alike
- moves the WiFi ladder into RefreshInterval, putting the week boundary under the native tests
- repairs the on-device integration suite, which hasn't compiled since connect() started returning WifiConnectionResult, and adds a compile-only run of it to CI

A week covers 98.2% of the check-in gaps devices actually recover from, and hourly wakes over that week cost about a quarter of a normal week's battery at the median 15 minute refresh.

Verified on an ESP32-C3: with a stale cached BSSID the fallback failed instantly before and connects after, and with the network also absent it reports failure on the timeout rather than inheriting the fast connect's answer.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,6 +72,14 @@ jobs:
       #- name: build (TRMNL_X)
       #  run: pio run -e TRMNL_X -j ${{ runner.os == 'Windows' && '1' || '4' }} -v
 
+      # Compile only. The suite needs a board to run, so nothing else here catches it drifting
+      # out of sync with the code it tests.
+      - name: build (on-device integration suite)
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          cp test/integration/test_config.h.example test/integration/test_config.h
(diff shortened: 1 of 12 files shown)


schrockwell:
For now, we are going with #557 as the quick fix to stop the bleeding. Then we can revisit fast reconnect with this patch.

schrockwell:
This is good stuff, but closing in favor of some more focused approaches that are based around these fixes.

PR #560, opened 2026-08-18
Spectra 6 faster refresh

ryanckulp:
this PR shaves 5s off my E1002 refresh timing. it may also increase panel's temperature sensitivity. according to some literature, if device environment is < 5°C or > 22°C this may create ghosting.


bitbank2:
I removed these files from the latest build because they are not needed. I will add this change to bb_epaper where it belongs. You can close this PR.

PR #561, opened 2026-08-18
Fixes portrait mode image problem on X

bitbank2:
The PNG decode logic was quitting the decode after 'height' lines were decoded. This logic needed to adapt to when the image is 90-degrees rotated and the height is actually the width.


ucffool:
Looking forward to this going into v1.8.14

schrockwell:
LGTM

Issue #513, opened 2026-07-26
E1002: v1.8.11 fixes color (#458) but fails to build for seeed_reTerminal_E1002 — missing <stdint.h>


empyreandance:
# E1002: blue and green render as white/yellow — intermittent regression of #308

**Device:** Seeed reTerminal E1002 (E Ink Spectra 6)
**Firmware:** `<1.8>` (flashed via https://trmnl.com/flash)
**Plugin:** private plugin, polling, markup is a single `<img>` of a server-rendered PNG

## Summary

Blue and green are not rendered correctly on the E1002 — they come out as white/yellow.
It is **intermittent**: the *same* source image renders correctly on one refresh and
incorrectly on the next, with no change to the image or the plugin.

This is the same symptom described in #308 (fixed in 1.7.7, closed 2026-06-05), but it is
occurring again on newer firmware, so filing separately as a regression rather than
commenting on the closed issue.

Quoting #308 for comparison — this matches exactly what I'm seeing:

> blue and green are rendered as white and yellow. This is not 100% reproducible:
> sometimes the colors are correctly shown ... three images render correctly, then the
> device freezes, and after a reboot blue and green are incorrect again.

## Evidence

I have **two photos of the same screen from the same source image** — one refresh rendered
correctly, the next did not. Attached below.

# Source Image sent to device 
<img width="800" height="480" alt="Image" src="https://github.com/user-attachments/assets/9a1645b5-d92b-4aae-a13f-90d1c61981d8" />

# Incorrent rendering
<img width="4032" height="3024" alt="Image" src="https://github.com/user-attachments/assets/568583ed-89cc-4c69-993d-e7cdcd51abf2" />

# Correct rendering
<img width="4032" height="3024" alt="Image" src="https://github.com/user-attachments/assets/e8fc3aef-a5b9-49a2-b1d0-c6bf87a778fd" />

**The source image is verifiably clean.** It is generated server-side and flattened to the
six Spectra primaries with **no dithering**, so there is nothing ambiguous for the firmware
to resolve:

```
800x480, exactly 6 distinct colors:
  white   rgb(255,255,255)  78.22%
  black   rgb(0,0,0)        14.99%
  green   rgb(0,255,0)       2.37%   <-- affected
  red     rgb(255,0,0)       1.92%
  blue    rgb(0,0,255)       1.27%   <-- affected
  yellow  rgb(255,255,0)     1.22%
```

[...]

PR #562, opened 2026-08-19
Fix the E1002 build by including stdint.h in spectra6.h

kk1987:
`src/displays/spectra6.h` uses `uint8_t` without including anything, and `spectra6.cpp` includes it before `Arduino.h`, so `pio run -e seeed_reTerminal_E1002` fails with `'uint8_t' does not name a type`. It's broken on main today.

This is the build failure from the title of #513. I wrote "Refs" rather than "Fixes" on purpose. The issue body is about the color regression, which #458 already fixed, and the comments also note that no 1.8.11 E1002 binary was ever published, which a code PR can't address.

### DIFF
--- a/src/displays/spectra6.h
+++ b/src/displays/spectra6.h
@@ -7,6 +7,8 @@
 
 #pragma once
 
+#include <stdint.h>
+
 /**
  * @brief Initializes the SPI interface for the Spectra6 display.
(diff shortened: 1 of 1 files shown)


ryanckulp:
implemented here as well, being fixed upstream so closing: https://github.com/usetrmnl/trmnl-firmware/pull/560

PR #565, opened 2026-08-19
Initialize the serial console on boards whose console is a plain UART

kk1987:
Since 67a3efb, `Serial.begin()` and `Log.begin()` only run under `DEV_FIRMWARE`. Boards with a USB-CDC console still print anyway, but the Seeed reTerminal E-series routes its console to UART0 through an on-board USB-UART bridge, so release builds are completely silent. The log strings are compiled into the binary. Logging is just never initialized.

That silence has a real diagnostic cost. The "no logs on 1.8.10 vs lots on 1.8.0" observation in #502 is this bug, not part of the display problem, and it sent that investigation sideways.

This adds a `UART_CONSOLE` opt-in flag and sets it for the E1001 and E1002 envs. I didn't gate on `ARDUINO_USB_CDC_ON_BOOT == 0` because envs that leave the macro undefined, the OG among them, default to 0 in the core, and their release behavior shouldn't change here.

On my E1001, a v1.8.12 build with the console initialized logs fine on UART0 at 115200. The `trmnl` env still compiles, and nothing changes without the flag.


GoReNeY:
That case was covered by #629, so no point in merging

PR #566, opened 2026-08-19
Compile the BYOD device envs in CI

kk1987:
Nothing in CI compiles the BYOD envs that `scripts/prepare_release.sh` cuts. The `pio run` step builds only the default `trmnl` env, and the publish workflows cover trmnl, TRMNL_X, and trmnl_4clr. So compile breakage in those envs ships silently. The E1002 build failure from the title of #513 is still present on main today and is a live example.

This adds one compile-only ubuntu job for the five BYOD envs that share the default espressif32 platform, with the same pio and ccache caching as the existing job. TRMNL_X_E1003 is left out because it needs the pioarduino platform and a much heavier job. All five envs build locally on this branch, so the gate starts green.

The branch carries the one-line stdint fix from #562 so E1002 can be in the list. It's the same patch, so git resolves it whichever lands first.


ryanckulp:
FYI for Rockwell, we're fixing `#include <stdint.h>` upstream following [this commit](https://github.com/usetrmnl/trmnl-firmware/pull/560/changes#diff-7be97fe133ecd3e0270b68a3fcc4c00edbabcc1c5db3e47342f05973d1202c8fR10).

schrockwell:
I merged in main after a major refactor. Will finish addressing this PR soon.

PR #567, opened 2026-08-19
Fix BWRY battery reading: measure voltage before turning on panel

schrockwell:
While testing, I found that my BWRY would report 2.4V even though it was fully charged or on USB power. This would occur during manual refreshes (button press).

I moved the voltage measurement ahead of the ePaper update, which fixed the issue. I suspect the voltage droops during the refresh.


ucffool:
good find, there are so few out there it would have dogged us for a long time.

PR #573, opened 2026-08-21
regression: `BYPASS_BQ27427_SOC` not handled

ronakjain90:
This PR makes changes so that every consumer of `battery().readSoc();` gets the same battery calculation logic. Previously `stateOfCharge` which determines the battery percentage was being read from `battery().readSoc();` which  returned the raw gauge value and ignored `BYPASS_BQ27427_SOC` entirely. 

This PR
1. moves the `BYPASS_BQ27427_SOC` logic from `display.cpp` to battery interface.
2. `gaugeInit()` now runs before the logo screen in setup, so the icon has a snapshot to read.

### DIFF
--- a/include/battery.h
+++ b/include/battery.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <battery_soc.h>
 #include <stdint.h>
 
 //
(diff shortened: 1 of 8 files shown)


schrockwell:
Thanks, but closing in favor of #574

PR #577, opened 2026-08-22
Store static IP enable state on edit

dominicdoty:
I ran into an issue on my device where I had an old static IP configuration saved in my browser local storage from a previous time setting up the device.

When I went to disable static IP by clicking the checkbox, all fields and the save/clear button disappear, but when I configured the device it was still set to static IP.

It appears that the design intent was for the user to click "Save" after unchecking the IP, but that button is hidden.

This pull request covers one of the possible ways to mitigate this - persisting the checkbox value to local storage immediately on edit.
Alternatively we could choose to display the "Save" and "Clear" buttons regardless of if the checkbox is checked.

Let me know if you'd prefer another approach.


ryanckulp:
thanks for this, but i think we should wait for #543 to merge first. need to double check but it may also fix this.

dominicdoty:
Ah shoot, yes that does look like it will do it. 
Sorry, should have checked before I fired this off last night, I just figured I'd fix it quick since it messed me up for a while yesterday.
I'll go ahead and close. Looking forward to the new portal, it looks nice!

PR #589, opened 2026-08-25
Fix status-202 (no content): stop false 'unable to connect' and always show the pairing code

joelhawksley:
## Problem

For a device whose MAC maps to an unclaimed/no-content record, `/api/display` returns **HTTP 200** with body `{"status":202}` (no `image_url`). Two firmware issues surfaced from this:

1. **Misleading "unable to connect".** `downloadAndShow()` still called `withHttp("")`; `https.begin("")` fails to parse the empty URL and the path reported `HTTPS_UNABLE_TO_CONNECT` ("unable connect to API/internet") even though WiFi, DNS, TLS and the API request all succeeded.
2. **Pairing code never (re)shown.** The pairing screen only rendered when the RTC `need_to_refresh_display` flag was `1`. That flag is zeroed after the first paint and after content, so a 202 arriving on a timer wake left a previously-paired device (whose server record was removed) silently fast-polling on stale content, never surfacing its current pairing code.

## Change (firmware-only)

- **Empty-URL guard:** skip the image fetch when there is no new image (`!status` or empty URL) and return the already-handled result instead of attempting `withHttp("")`.
- **Always land on the pairing code:** on `HTTPS_NO_REGISTER`, render the pairing-code screen regardless of `need_to_refresh_display`, deduped against the last painted code (kept in RTC, cleared when content is shown) so re-pairing after a server-side deletion surfaces the code without e-ink thrash on fast polls. The decision is extracted into `shouldShowPairingCode()` (`lib/trmnl/include/pairing.h`) and unit-tested.

## Tests

- `test_parse_api_display`: added cases for `{"status":202}` and `status:0` with empty `image_url` (must parse to no image URL).
- `test_pairing`: unit tests for `shouldShowPairingCode()` (registered vs. unregistered, force-refresh, code-changed, same-code dedupe).
- `pio test -e native` green (80 cases).

## Verified on hardware (reTerminal E1001)

```
status: 202
No image URL to download (status=0, url_len=0); skipping image fetch
request result - HTTPS_NO_REGISTER
display_show_msg2 start / friendly id case   -> pairing code 598364 shown
```

Pairing screen repainted once across two 202 polls (deduped); no "unable to connect". Regression-built `trmnl`, `seeed_reTerminal_E1001`, and `TRMNL_X_E1003`.

### DIFF
--- a/lib/trmnl/include/pairing.h
+++ b/lib/trmnl/include/pairing.h
@@ -1,0 +1,17 @@
+#pragma once
+
+#include <string.h>
+
+// Decide whether to (re)paint the pairing-code screen for a NO_REGISTER
+// (/api/display HTTP 200 body status 202, or otherwise unregistered) response.
+//
+// Repaint when the device is unregistered AND either a repaint is forced
+// (power-on, or content just cleared the last-shown tracker) or the stored
+// friendly id differs from the code we last painted. This surfaces the pairing
+// code after a server-side device deletion without e-ink thrash on fast polls.
+inline bool shouldShowPairingCode(bool no_register, bool force_refresh, const char *stored_friendly_id,
+                                  const char *last_shown_id) {
+  if (!no_register) return false;
+  if (force_refresh) return true;
+  return strcmp(stored_friendly_id ? stored_friendly_id : "", last_shown_id ? last_shown_id : "") != 0;
+}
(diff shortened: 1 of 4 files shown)


schrockwell:
Jinx! We found the same issue and fixed it in #605, which was just released in v1.8.16. Thanks for your contribution.

PR #604, opened 2026-08-27
Fix time sync fail due to timeout

ronakjain90:
**Problem:**
I noticed that my TRMNL X has lost sync of time, it was ahead of the time by 10 minutes. To investigate, I tried to bypass the 24 hrs check and the clock didn't sync even then. I found that log consistently succeeded in 5000ms (I added some diagnostics logs)

**before**
logs with diagnostics:
```
I: src/misc/clock/clock.cpp [78]: Time synchronization...
I: src/misc/clock/clock.cpp [105]: Time synchronization succeed after 5000 ms
```
logs from main, Current time is wrong and is off by 30 mins.
```
I: src/misc/clock/clock.cpp [58]: Time synchronization...
I: src/misc/clock/clock.cpp [63]: Time synchronization succeed!
I: src/misc/clock/clock.cpp [69]: Current time - Mon Aug 31 14:48:13 2026
```

I found that by default [getLocalTime](https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/Arduino.h#L273) has a timeout of 5s. hence it silently succeeds and the device never attempts to sync the time for next 24 hrs.

**Solution:**
This PR sets `esp_sntp_set_sync_status` to `SNTP_SYNC_STATUS_RESET` and then polls the status until it changes to `SNTP_SYNC_STATUS_COMPLETED` for 10s.

after logs:
```
I: src/misc/clock/clock.cpp [38]: Using NTP: time.google.com, fallback: pool.ntp.org
I: src/misc/clock/clock.cpp [58]: SNTP server[0]: time.google.com
I: src/misc/clock/clock.cpp [62]: Time sync started
I: src/misc/clock/clock.cpp [74]: Time sync succeed in 2200 ms
I: src/misc/clock/clock.cpp [80]: Current time - Thu Aug 27 13:47:48 2026
```

### DIFF
--- a/src/bl.cpp
+++ b/src/bl.cpp
@@ -1184,13 +1184,10 @@ void bl_init(void)
 #endif
 
   // clock synchronization
-  if (systemClock().setTimeFromNTP())
-  {
+  if (systemClock().setTimeFromNTP()) {
     time_since_sleep = preferences.getUInt(PREFERENCES_LAST_SLEEP_TIME, 0);
-    time_since_sleep = time_since_sleep ? systemClock().getTime() - time_since_sleep : 0; // may be can be used even if no sync
-  }
-  else
-  {
+    time_since_sleep = time_since_sleep ? systemClock().getTime() - time_since_sleep : 0; // maybe can be used even if no sync
+  } else {
     time_since_sleep = 0;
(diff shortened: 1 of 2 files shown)


stevekennedyuk:
I’d also hope it uses local NTP servers in DHCP options

PR #605, opened 2026-08-27
Fixed the setup screen being replaced by an API error

ikraamg:
A device that has finished /api/setup but has not been claimed to an account yet is answered with body status 202, and that response carries no image_url. The 202 arm of `handleApiDisplayResponse` never touches the global `filename` buffer, so it is still empty when `downloadAndShow` reaches the image fetch, which runs regardless. `HTTPClient::begin` rejects the empty URL, and since #570 that failure is assigned back over `result`, so `HTTPS_NO_REGISTER` becomes `HTTPS_UNABLE_TO_CONNECT` and the friendly ID screen is replaced by "WiFi connected, unable connect to API."

- returns early from the image fetch when there is no URL to fetch
- adds a native test covering the production 202 body and the URL check `HTTPClient::begin` runs

`result = withHttp(` is absent from v1.8.11 through v1.8.14 and present in v1.8.15 and on main, so only 1.8.15 devices are affected. It only bites in the window between setup and the device being claimed: after 24 hours the server answers those devices with a body-0 setup screen instead, and claiming the device ends it too, which is why it looks like it resolves on its own.

In production telemetry the "unable to connect" log line went from about one a day to 416 on 26 Aug and 788 on 27 Aug, and 1,205 of those 1,211 lines came from 1.8.15. On 27 Aug 42 devices on 1.8.15 saw a 202 and 40 logged the error. Affected models are TRMNL X and OG 2-bit. One device logged 48 of them over 40 minutes, almost all woken by the touch bar, so each retry repaints the error screen with a full refresh.

Built `pio run -e trmnl` and ran `pio test -e native -f test_unclaimed_device_response`. `TRMNL_X` does not build on my machine (missing `fatfs` python module in the espressif32 builder) and fails the same way on an unmodified checkout, so that one still needs a build from someone whose toolchain is complete.

### DIFF
--- a/src/bl.cpp
+++ b/src/bl.cpp
@@ -1649,6 +1649,11 @@ static https_request_err_e downloadAndShow()
   }
 #endif // BOARD_TRMNL_X
 
+  // A 202 carries no image_url, so filename is still empty here. HTTPClient::begin() rejects
+  // that and the fetch below would report it as an API failure over the friendly ID screen.
+  if (filename[0] == '\0')
+    return result;
+
   result = withHttp(
       filename,
       [&](HTTPClient *httpsp, HttpError error) -> https_request_err_e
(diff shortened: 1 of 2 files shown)


schrockwell:
Tested and confirmed working on X. Thanks!!

PR #607, opened 2026-08-28
Report USB and charging status on reTerminal E1001/E1003

joelhawksley:
## Summary

Report USB connection and battery charging status for reTerminal E1001 and E1003 through the existing API headers.

## Testing

- Built `seeed_reTerminal_E1001`
- Built `TRMNL_X_E1003`

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -143,6 +143,10 @@
 #define VBAT_SWITCH_LEVEL HIGH // load switch enable pin active level
 #define PIN_BUZZER        45          // passive buzzer (MLT-8530) via NPN transistor
 #define BUZZER_FREQ       2700       // resonant frequency of MLT-8530 in Hz
+#define CHARGER_SY6974               // SY6974B I2C charger PMIC
+#define CHARGER_I2C_BUS   Wire1      // SY6974B sits on I2C1
+#define CHARGER_I2C_SDA   39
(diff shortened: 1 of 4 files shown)


schrockwell:
Hi again @joelhawksley. We are moving away from #ifdef-based device definitions, so this needs some redesign. @bitbank2 do you have any suggestions on how to handle this? Maybe new columns in the device table?

bitbank2:
Thanks for the PR @joelhawksley. @schrockwell We can add this functionality without new columns for the device definition list. The existing columns for SDA/SCL are all that's needed to support the E1001/E1003 functionality. I have a E1001 on order, but Seeed says it won't ship until Sept 18th. Until then, I can add this new code to my devices_simplified_2 branch and Joel can see if it works correctly.

bitbank2:
Update: I see that the E1004 (which I have) also uses the same SY6974B PMIC, so I can add and test this new code this week.

schrockwell:
Great, thanks @bitbank2. We just merged `devices_simplified_2` into main, there's no need to continue on that branch, and it may cause conflicts it diverges. So please make a new branch and PR, and we can supersede this one.

joelhawksley:
Great, I will move this back to draft and confirm y'alls fix is sufficient once it lands before closing.

PR #611, opened 2026-08-29
Create a dedicated image download error screen page.

ronakjain90:
**Problem:**
When the API results in success, but the Image download results in failure. The device shows `"WiFi connected, request to API failed."`. This statement is incorrect as the API call itself was successful, it's having trouble to download image, which could have failed for various unknown reasons.

The log sent to the server says `"[HTTPS] GET... failed"` which unfortunately doesn't give support team / customers right feedback about the image server being in patchy / degraded state.

**Solution:**
- Create a dedicated error screen when Image download fails and display the relevant image download error screen. This is especially useful for BYOS / Alias / Redirect plugins.
- The https error log is now changed to `[HTTPS] Modem Image GET... failed` and `[HTTPS] Image GET... failed` respectively.

**Tests:**
- [X] TRMNL X
- [ ] TRMNL OG
- [ ] TRMNL BWRY


schrockwell:
This is a nice, necessary improvement. I am going to rework this to keep the screen but change the image fetch logic to have some retries before showing the new error message.

schrockwell:
Cherry-picked this to #635 - will apply co-author credit!

PR #613, opened 2026-08-31
Move HTTPS error log names to `types.h`

ronakjain90:
**Problem:**
While working on #611, I noticed that the `https_request_err_e` and `szHTTPErrors` were out of sync. The later doesn't have `HTTPS_TIMED_OUT` string. This happens when the programmer doesn't know that an entry has to be added at 2 places.

**Solution:**
In this PR we 
- Remove the duplicate `szHTTPErrors[]` array from `bl.cpp` and get the error log names via `https_request_err_str()` in `types.h`.
- Added the earlier missing `HTTPS_TIMED_OUT`

**Tests:**
- [X] TRMNL X
- [ ] TRMNL OG
- [ ] TRMNL BWRY

### DIFF
--- a/include/types.h
+++ b/include/types.h
@@ -29,4 +29,36 @@ enum https_request_err_e {
   HTTPS_TIMED_OUT
 };
 
+struct https_request_err_name {
+  https_request_err_e code;
+  const char *name;
+};
+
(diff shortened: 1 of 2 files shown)


schrockwell:
Nice refactor, clean and effective. Thanks!

PR #617, opened 2026-09-02
Mount LittleFS on parallel ePaper boards

ronakjain90:
`BOARD_X_CLASS` is no longer used since #550


schrockwell:
Great find, thanks. I confirmed there are no other references to `BOARD_X_CLASS` after this.

PR #618, opened 2026-09-02
Introducing Client draw

ronakjain90:
Introducing Client Draw. Device capable of partial refresh (like TRMNL X) would be able to make use of Client Draw plugins (like Clock).

**How it works ?**
This works by FW reading the metadata from png files, which then makes use of partial refresh capability to draw new content programatically without needing to call API services. Plugins that can live in isolation can make use of this Client Draw feature.

**png-metadata data structure**
Uses JSON data to allow using unstructured data, other format such as protobuf/csv etc were explored, JSON was preferred over flexibility.
```json
[
  {
    "v": 1,
    "name": "clock",
    "rect": [
      0,
      0,
      400,
      240
    ],
    "type": "analog/digital",
    "tz": 0
  }
]
```

**What size can it support?**
Client draw can support full-screen / mashups comfortably. The server determines the rectangle grid where the client has to draw its content (refer to `rect` above). The client chooses the font size dynamically based on the available width.


bitbank2:
Sorry @ronakjain90 I just took a look at this today. This code is 99% unnecessary and just plain bad.
The implementation is one big hack that we should avoid.
The hacking at FastEPD and the PNG parsing is especially awful and not needed.
The only part that is reasonable to save is placing JSON in a PNG tEXt chunk; toss all of the rest.

The PNG with metadata can be created on the back-end and FastEPD and PNGdec can ALREADY work with that data without any patches/hacks.

PR #619, opened 2026-09-02
Return unknown for unlisted WiFi status

ronakjain90:
I got a crash on my local when WiFi has status code `WL_STOPPED` which is missing in the list. [reference](https://github.com/espressif/arduino-esp32/blob/6048a624f084ea7f645384fd91c57f43e633875d/libraries/WiFi/src/WiFiType.h#L45)


schrockwell:
Thanks. I also added a `WL_STOPPED` handler.

PR #621, opened 2026-09-03
Fix gas gauge current polarity, and add temporary gas gauge headers

GoReNeY:


schrockwell:
Formatted and tested on X with an HTTP proxy. Looks great, thank you. Will work on getting this into core next.

```
    Gauge-Capacity: 23020/23410
    Gauge-Health: 100
    Gauge-SOC: 99
```

PR #625, opened 2026-09-05
Adds support for the XTEINK X3

bitbank2:
This change adds support for the XTEINK X3. We need to also create a new custom device/size for it (792x528)


bitbank2:
fixed and merged into main

PR #638, opened 2026-09-12
Added new logic to purge old playlist items from FLASH

bitbank2:
This code supports the backend change which shares the current playlist items in each API response. This list is compared against cached images and those that are no longer on the playlist are deleted. This logic prevents old/removed playlist items from accidentally being displayed on the X when the user navigates forward or back with the touchbar.

### DIFF
--- a/src/filesystem.cpp
+++ b/src/filesystem.cpp
@@ -2,7 +2,7 @@
 #include <filesystem.h>
 #include <inttypes.h>
 #include <trmnl_log.h>
-
+#include <api_types.h>
 #if defined(PARALLEL_EPD)
 #include <LittleFS.h>
 #define FS LittleFS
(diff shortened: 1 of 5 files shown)


src/filesystem.cpp
```diff
@@ -128,22 +177,14 @@ void filesystem_purge_old_file(const char *name) {
...
+ 
     strcpy(szTemp, "/"); // needed on this file operation
     strcat(szTemp, file.name());
 
     Log_info("Comparing name %s with %s, timestamp %" PRIu32 ", current time %" PRIu32, name, file.name(), timestamp,
              (uint32_t)now);
     if (strncmp(name, szTemp, 14) == 0) { // older version of the same file
       Log_info("Deleting older version of plugin image %s - %s", name, file.name());
-      bDel = true;
-    } else if ((uint32_t)now > timestamp + 60 * 60 * 24) { // More than 24h old, or no timestamp
```
ronakjain90, line 141: We have to keep this logic @bitbank2, as otherwise the core server doesn't know the file name of `redirect` plugin and the SPIFF storage will keep growing thus never having a chance to recover. 

This change will also brick all the devices with BYOS implementation.

bitbank2: Sorry, I don't understand. The new logic will always delete older files and those that aren't in the current playlist. The FLASH will never fill up.

ronakjain90: We are assuming that server changes will be ported over by all the BYOS implementations from day 1, this is a breaking change and we need to handle the BYOS + redirect plugin implementation gracefully.

bitbank2: Hmm, it seems important enough to port it to BYOS. Hopefully not a big change to include the playlist in the API response. Any other way of managing the local copies of plugin images would allow for running out of storage.

PR #643, opened 2026-09-16
Moved the stored-log upload after the screen refresh

ikraamg:
Stored logs were posted to /api/log twice before the image was drawn: once before the /api/display check-in and once between the image download and the refresh. Each post waits up to 15 s for the server, so on 2026-09-16, when /api/log accepted connections and answered nothing for 8 minutes, every device holding a stored error log showed its screen 17 to 35 s late (about 870 slow wakes in 20 minutes in the fleet telemetry). The upload now runs once, after display_show_image() returns and before WiFi is turned off, so nothing the log endpoint does can delay the screen. Failure handling is unchanged: an unsent log stays in NVS and is retried on the next wake, the call after downloadAndShow() is kept because it ships the logs before a server-ordered reset wipes NVS, and goToSleep() still posts on the paths that never reach a draw. The only cost is that WiFi stays associated through the panel refresh, which the cached-image path already does today.

Tested on an OG against scripts/mock_server.py (#642) with /api/log hanging 20 s. On main a timer wake went log post (15 s), check-in, image, log post (15 s), draw: about 15 s from image in hand to screen. On this branch the same wake went check-in, image, draw, log post: about 3 s from image in hand to screen, and the wake starts with the check-in instead of a 15 s stall. With a healthy server the stored logs were posted 3 s after the draw, answered 204 and cleared, and the next wake had nothing to post. Both env:trmnl and env:TRMNL_X build.


ronakjain90:
sending the stored logs after display would result in WiFi chip remaining powered on for the entire duration of the screen refresh time, thus consuming more battery. this change will impact color EPDs more.

PR #649, opened 2026-09-17
Read and report panel revision

schrockwell:
Added function to read the EPD panel revision as a 32-bit value (panel id) to be used to identify the correct temperature profile

Also reported via `/api/setup` and `/api/display` in the new `Panel-Rev` header:

- [x] OG
- [x] X - no header 

```
>>> GET /api/setup  from 192.168.40.28
    Host: 192.168.40.27:8888
    User-Agent: ESP32HTTPClient
    Connection: close
    Accept-Encoding: identity;q=1,chunked;q=0.1,*;q=0
    ID: D8:3B:DA:F5:28:3C
    Content-Type: application/json
    FW-Version: 1.8.16
    Model: og
    Panel-Rev: fdfffe19
```

### DIFF
--- a/src/display.cpp
+++ b/src/display.cpp
@@ -153,6 +153,70 @@ static bool display_update_epaper(int refreshMode, bool wait, bool writePlane = 
 }
 #endif
 
+static void WriteSPIByte(uint8_t data)
+{
+  for (int i=0; i<8; i++) {
+    digitalWrite(EPD_MOSI_PIN, (data & 0x80) ? 1:0);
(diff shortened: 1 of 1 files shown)


schrockwell:
I renamed this concept from "panel ID" to "panel revision" (with header `Panel-Rev`) because:

1. the SPI command is called "REV"
2. in core, we have talked about adding a `Panel` model, which would necessarily have a foreign key called `panel_id`, which would be confusing, so I named that column `panel_revision` instead

